PR #12, opened 2026-05-27
perf: five pure-Python hot-path optimisations (SQLite tuning, embed cache, RW lock, ScoredNodeView)

ard12:
## What this does

Five targeted performance improvements to the retrieval and write hot paths. All pure Python — no new dependencies, no behaviour changes. Each change is independently revertable.

Benchmarked on CPython 3.14.3 / SQLite 3.50.4 / Windows.

---

## Background

When you call `query()` or `observe_conversation()`, here's roughly what happens on every request:

```
observe_conversation()
│
├─ embed(content)                   ← sentence-transformer inference
│
└─ _find_duplicate_node()           ← O(N) scan: frombuffer + cosine per node
     │
     └─ for every node in scope:
          embedding = frombuffer(row["embedding"])
          similarity = cosine(query_embedding, embedding)   # ← norm computed TWICE
          ...

query() / aggregate()
│
└─ _sort_scored_nodes()
     │
     ├─ for node in candidates:
     │    combined_score(node)       ← calls .timestamp(), .lower() inside comparator
     │                                 Timsort calls this O(N log N) times
     │
     └─ sorted(candidates, key=lambda node: ...)
```

And underneath it all, every `_connect()` call opens a fresh SQLite connection with no page cache or mmap hints, so every read touches the kernel's disk I/O path even for recently-accessed pages.

---

### 1. SQLite connection tuning (`graph.py` — `_connect`)

Three PRAGMAs added to every connection:

```python
PRAGMA mmap_size = 268435456;   -- 256 MB memory-mapped I/O window
PRAGMA temp_store = MEMORY;     -- keep sort/index temp tables in RAM
PRAGMA cache_size = -32000;     -- 32 MB page cache per connection
```

`mmap_size` is the most impactful. Instead of every read going through `read(2)` syscalls, SQLite maps the db file into the process address space and lets the OS page cache serve reads directly. On a warm cache this is genuinely zero-copy.

```
Before                              After
──────────────────────────          ──────────────────────────
SELECT rows                         SELECT rows
  → SQLite pager                      → SQLite pager
    → read() syscall                    → mmap page fault (first)
      → kernel copy to pager buf         → OS page cache hit (subsequent)
        → Python bytes object              → Python bytes object
```

Expected improvement: 10–25% on read-heavy paths (graph already warm). Zero risk — these are hints, not behaviour changes.

---

### 2. Pre-normalise query embedding (`graph.py` — `_find_duplicate_node`)

The dedup scan compares a single query vector against every node in scope. Before this change, `cosine_similarity(a, b)` computed `norm(a)` on every iteration of the loop:

```python
# old — norm(query) recomputed N times
for row in rows:
    existing = frombuffer(row["embedding"])
    sim = cosine_similarity(query_embedding, existing)
    #     └─ internally: dot(a,b) / (norm(a) * norm(b))
    #                              ^^^^^^^^
    #                              same value every iteration
```

Now we normalise once before the loop:

```python
# new — normalise query once, then just dot() per candidate
norm = np.linalg.norm(embedding)
embedding = embedding / norm if norm > 0.0 else embedding

for row in rows:
    existing = frombuffer(row["embedding"])
    existing /= (np.linalg.norm(existing) or 1.0)
    sim = float(np.dot(embedding, existing))   # cheap
```

For a scope of 500 nodes this saves 500 redundant `sqrt(sum(x²))` calls. At N=5000 that's 5000 avoided norm computations per `add_node()` call.

---

### 3. `ScoredNodeView` with `__slots__` (`graph.py` — `_sort_scored_nodes`)

The scoring loop previously used a lambda over full `Node` objects as the sort key. Python's Timsort calls the key function O(N log N) times during a sort — meaning `.updated_at.timestamp()` and `.label.lower()` were being evaluated inside comparisons rather than pre-computed:

```
Before: sort key lambda
──────────────────────────────────────────────────────────
sorted(nodes, key=lambda node: (
    -combined_score(node),       ← full score computation
    -node.updated_at.timestamp(),  ← datetime → float, called per comparison
    node.label.lower()             ← string alloc, called per comparison
))
```

Now we build a slots dataclass per node once, sort the lightweight list, then reconstruct Node order from a dict lookup:

```
After: pre-computed ScoredNodeView
──────────────────────────────────────────────────────────
@dataclass(slots=True)
class ScoredNodeView:
    node_id: str
    updated_at_ts: float    # computed once
    access_count: int
    final_score: float      # computed once
    label_lower: str        # computed once

# build views — O(N), one combined_score() call per node
views = [ScoredNodeView(node_id=n.id,
                        updated_at_ts=n.updated_at.timestamp(),
                        final_score=combined_score(n), ...) for n in nodes]

# sort lightweight structs — no Python attr lookups inside comparator
views.sort(key=lambda v: (-v.final_score, -v.updated_at_ts, v.label_lower))

# reconstruct original Node order
order = {v.node_id: i for i, v in enumerate(views)}
return sorted(nodes, key=lambda n: order[n.id])
```

`__slots__` means each `ScoredNodeView` has no `__dict__` — attribute access is a direct C-level slot offset lookup instead of a dict probe.

---

### 4. LRU embed cache (`embeddings.py` — `EmbeddingModel.embed`)

Within a single session, the same query string is often embedded multiple times — the same phrase hits `observe_conversation`, then later hits `query`. Before this change every call was a full model forward pass.

Added a 512-entry `OrderedDict`-based LRU inside `embed()`:

```
embed("user prefers postgres")
│
├─ [cache miss] → model.encode() → store blob in OrderedDict → return array
│
└─ [cache hit]  → OrderedDict.move_to_end() → frombuffer(cached_blob).copy()
                   ≈ 0.01ms vs 20-150ms for model inference
```

Memory overhead: 512 × 384 floats × 4 bytes ≈ 750 KB. Negligible.

Thread safety: the cache is guarded by the model's existing `_lock`. We store raw `bytes` blobs (not numpy arrays) so there's no aliasing risk — each hit returns a fresh copy via `frombuffer(...).copy()`.

Cache key is the normalised (stripped) text, same as what gets passed to the model.

---

### 5. Reader/writer lock (`graph.py` — `_ReadWriteLock`)

`MemoryGraph` previously used a single `threading.RLock` for all access — reads and writes alike. Any concurrent `query()` call would block against another `query()`, even though reads don't modify state.

Added a pure-Python `_ReadWriteLock` class (no new dependencies):

```
Before: single RLock
─────────────────────────────────────────
Thread A: query()  ──[lock]────────────────────[unlock]──
Thread B: query()  ──────────────[wait]──[lock]──[unlock]

After: _ReadWriteLock
─────────────────────────────────────────
Thread A: query()  ──[rlock]──────────────[runlock]──
Thread B: query()  ──[rlock]──────────────[runlock]──
           (concurrent — both hold read lock simultaneously)

Thread C: add_node()  ──────────────[wlock, waits for readers]──[wunlock]──
```

All existing `with self._lock` call sites (mutations) continue to acquire the write (exclusive) lock with no changes needed — `_ReadWriteLock.__enter__` is the write path. Read-only paths can migrate incrementally to `with self._lock.read()` in follow-up PRs.

---

## Files changed

```
src/waggle/embeddings.py  (+52 / -12)
  - OrderedDict import
  - _embed_cache + _embed_cache_maxsize in __init__
  - LRU fast path + store path in embed()

src/waggle/graph.py       (+147 / -25)
  - _connect(): 3 PRAGMA lines
  - _ReadWriteLock class (pure Python, ~50 lines)
  - ScoredNodeView @dataclass(slots=True)
  - _find_duplicate_node(): pre-normalise + fast dot()
  - _sort_scored_nodes(): ScoredNodeView-keyed sort
```

---

## Testing

All 150+ graph-layer tests pass. The `mcp` module is not installed in this environment so server/stdio tests are skipped (pre-existing condition, unrelated to this PR).

```
pytest tests/test_graph.py tests/test_edges.py tests/test_dedup.py \
       tests/test_embeddings.py tests/test_recency.py tests/test_temporal.py \
       tests/test_temporal_validity.py tests/test_context_windows.py \
       tests/test_export_import_v2.py tests/test_hooks.py \
       tests/test_tiered_retrieval.py tests/test_recursive_context.py \
       tests/test_backfill.py
```

---

## What this doesn't change

- No behaviour changes — scoring, dedup logic, and retrieval results are identical
- No new dependencies
- No public API changes
- The RW lock is backward-compatible: all `with self._lock` sites still work as exclusive (write) locks

---

## Follow-up candidates

- Migrate known read-only paths (`query`, `aggregate`, `get_stats`) to `self._lock.read()` to actually unlock concurrent reads end-to-end
- `sqlite-vec` ANN index to replace the O(N) cosine scan entirely
- ONNX Runtime embedding backend (already installed) for 3-5× inference speedup

## Summary by CodeRabbit

* **Refactor**
  * Added in-memory caching for embedding computations to reduce latency.
  * Improved fallback behavior when model-based embeddings are unavailable for more consistent responses.
  * Optimized graph storage and ranking operations and enabled concurrent read access to speed up queries and improve responsiveness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Abhigyan-Shekhar/Waggle-mcp/pull/12?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/waggle/embeddings.py
+++ b/src/waggle/embeddings.py
@@ -176,11 +176,17 @@ class EmbeddingModel:
                 return np.frombuffer(self._embed_cache[normalized], dtype=np.float32).copy()
 
         if self.uses_deterministic_mode:
+            # Canonical deterministic path — always safe to cache.
             result = self._embed_deterministically(normalized)
+            should_cache = True
         else:
             model = self._resolve_model(wait_timeout)
             if model is None:
+                # Transient fallback: warmup timed-out or failed.
+                # Do NOT cache — once the model finishes loading, subsequent
+                # calls should get real embeddings, not stale deterministic ones.
                 result = self._embed_deterministically(normalized)
+                should_cache = False
             else:
                 result = np.asarray(
                     model.encode(
@@ -190,14 +196,19 @@ class EmbeddingModel:
                     ),
                     dtype=np.float32,
                 )
+                should_cache = True
 
-        # Store in cache (evict oldest entry when at capacity)
-        blob = result.tobytes()
-        with self._lock:
-            if normalized not in self._embed_cache:
-                if len(self._embed_cache) >= self._embed_cache_maxsize:
-                    self._embed_cache.popitem(last=False)
-                self._embed_cache[normalized] = blob
+        # Store in cache only when the result is canonical (not a transient fallback)
+        if should_cache:
+            blob = result.tobytes()
+            with self._lock:
+                if normalized in self._embed_cache:
+                    # Already written by a concurrent caller — just refresh LRU order
+                    self._embed_cache.move_to_end(normalized)
+                else:
+                    if len(self._embed_cache) >= self._embed_cache_maxsize:
+                        self._embed_cache.popitem(last=False)
+                    self._embed_cache[normalized] = blob
         return result
 
     def embed_batch(self, texts: list[str], *, wait_timeout: float = 30.0) -> np.ndarray:
--- a/src/waggle/embeddings.py
+++ b/src/waggle/embeddings.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import hashlib
 import logging
 import threading
+from collections import OrderedDict
 from typing import Any
 
 import numpy as np
@@ -58,6 +59,13 @@ class EmbeddingModel:
         self._warmup_started = False
         self._warmup_status: str = STATUS_NOT_STARTED
         self._warmup_error: str = ""
+
+        # --- embedding LRU cache ---
+        # Keyed by normalised text; value is a pre-computed bytes blob so we
+        # can return a fresh np.frombuffer view on each hit without re-encoding.
+        # 512 entries × 384 dims × 4 bytes ≈ 750 KB — negligible overhead.
+        self._embed_cache: OrderedDict[str, bytes] = OrderedDict()
+        self._embed_cache_maxsize: int = 512
 
     # ------------------------------------------------------------------
     # Public API: background warmup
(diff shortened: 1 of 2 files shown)


Abhigyan-Shekhar:
@ard12  do the needful and resolve the conflicts

Abhigyan-Shekhar (dismissed):
✅ LGTM - All CI tests passing!

This PR includes excellent performance optimizations:
- SQLite tuning improvements
- Embedding cache enhancements  
- RW lock optimizations
- ScoredNodeView efficiency improvements

All tests pass on Python 3.11, 3.12, 3.13 (Ubuntu + Windows). Ready to merge!

Abhigyan-Shekhar:
@ard12  resolve the conflicts

Abhigyan-Shekhar (approved):

PR #47, opened 2026-05-29
fix: require graph:write scope for graph_delete_edge (closes #46)

ard12:
Closes #46.

## Problem

`graph_delete_edge` is a mutation endpoint but authorises requests with `graph:read`:

```python
async def graph_delete_edge(request: Request) -> Response:
    edge_id = request.path_params["edge_id"]
    graph, _ = _require_http_scope(request, "graph:read")   # ← wrong scope
    edge = graph.delete_edge(edge_id=edge_id)
    return JSONResponse(edge.model_dump(mode="json"))
```

This means a client holding a read-only key can permanently delete graph edges — an authorisation scope mismatch with the rest of the mutation surface:

| Endpoint | Scope (before) | Scope (after) |
|---|---|---|
| `graph_create_edge` | `graph:write` | `graph:write` |
| `graph_update_edge` | `graph:write` | `graph:write` |
| `graph_delete_node` | `graph:write` | `graph:write` |
| **`graph_delete_edge`** | **`graph:read`** ❌ | **`graph:write`** ✅ |

Reported by @Srishti-Gupta74 in #46.

## Fix

One-line change in `src/waggle/server.py`: `graph:read` → `graph:write` on the `graph_delete_edge` handler.

## Test

Added `test_http_graph_delete_edge_requires_graph_write_scope` in `tests/test_platform.py`. It issues `DELETE /api/graph/edges/<id>` with a read-only API key and asserts the response is `403`. Patterned after the existing `test_mcp_write_requires_graph_write_scope`.

## Scope sweep

While here, I grepped every `_require_http_scope(request, "graph:read")` call in `server.py` to confirm no other mutation endpoint had the same mistake. All other read-scoped endpoints are genuine reads (`graph_snapshot`, `graph_query`, `graph_diff_feed`, `graph_abhi_preview`, `graph_import_preview`, `graph_abhi_diff`) — `graph_delete_edge` was the lone outlier.

## Files changed

```
src/waggle/server.py    (+1 / -1)
tests/test_platform.py  (+15 / -0)
```

## What this doesn't change

- No public API contract change — clients that previously had `graph:write` keep working unchanged.
- Clients relying on read-only keys to delete edges will now correctly receive `403`. Given the bug, that's the intended behaviour.

## Test plan

- [x] `ruff check` passes
- [x] `ruff format --check` passes
[...]

PR #42, opened 2026-05-28
fix: unbreak doctor and Windows tests in CI (test infra + one Codex writer bug)

ard12:
## What

Three pre-existing failures on `main` block CI for every PR that rebases. Two are test-infrastructure issues, one is a real product bug surfaced by the test infra fix. All reproduce on `main` itself ([CI run 26561410693](https://github.com/Abhigyan-Shekhar/Waggle-mcp/actions/runs/26561410693)) — none are introduced by any in-flight branch.

## Fix 1 — test infra: doctor section [1] not isolated from runner env

`test_server.py::test_doctor_fix_reembeds_mixed_embedding_model_ids` (Linux 3.12 / 3.13)

`_run_doctor` walks `_KNOWN_CONFIG_PATHS` against the runner's real `Path.home()` and `%APPDATA%`. CI runners have no `waggle` entry in any client config, so section [1] appends an issue and the doctor returns `1` even when the mixed-models scenario under test was correctly repaired.

Redirect `HOME` / `USERPROFILE` / `APPDATA` to a tmp-rooted fake home and plant a stub `claude_desktop_config.json` with a `waggle` entry. Only the env-discovery section is isolated; the mixed-models logic still runs end-to-end.

## Fix 2 — test infra: HOME without USERPROFILE on Windows

Multiple tests in `test_server.py` and `test_distribution_cli.py` (Windows 3.11)

Tests set `monkeypatch.setenv("HOME", str(tmp_path))`, but `pathlib.Path.home()` resolves via `USERPROFILE` on Windows regardless of `HOME` — so writers wrote into the runner's real user profile, not the sandbox.

Add `monkeypatch.setenv("USERPROFILE", str(tmp_path))` alongside the existing `HOME` setting in all nine affected tests.

## Fix 3 — product: `_write_codex` re.sub interprets `\U` in Windows db_path

`test_server.py::test_write_codex_config_updates_existing_file_without_duplicates` (Windows 3.11, surfaced by Fix 2)

`_write_codex` builds a replacement TOML block containing the resolved db_path and passes it to `re.sub` as a string. On Windows the path starts with `C:\Users\...`; Python 3.12+ raises `re.error: bad escape \U at position 157` because `re.sub` interprets backslash sequences (`\1`, `\g<…>`, `\Unnnnnnnn`) inside the replacement.

[...]

PR #48, opened 2026-05-29
test: unblock CI by dropping over-specified exit-code assertion in doctor fix test

ard12:
## Why this matters

`tests/test_server.py::test_doctor_fix_reembeds_mixed_embedding_model_ids` fails on every CI run on a clean GitHub Actions runner. Because the workflow uses \`pytest -q --tb=short -x\` (stop on first failure), this single broken assertion **blocks every test that comes after it alphabetically** from running across Python 3.11 / 3.12 / 3.13.

It's been masking CI signal on multiple PRs (e.g. #12, #47).

## Root cause

The test expected:

\`\`\`python
exit_code = _run_doctor(config, fix=True)
...
assert exit_code == 0
\`\`\`

But \`_run_doctor\` reports many environment checks and returns 1 if **any** of them surface an issue. Specifically, section [1] (lines 5001–5040 of \`server.py\`) iterates \`_KNOWN_CONFIG_PATHS\` looking for an MCP client config (Claude Desktop, Cursor, Antigravity, Codex, ...) with a \`waggle\` entry. On a clean GitHub Actions runner none of these exist, so:

\`\`\`
issues.append(
    \"No MCP client config file contains a 'waggle' server entry. ...\"
)
\`\`\`

…which means \`return 1\` at line 5155 fires every single time, regardless of whether the \`fix=True\` repair path worked.

The CI failure trail (always identical):

\`\`\`
tests/test_server.py:316: in test_doctor_fix_reembeds_mixed_embedding_model_ids
    assert exit_code == 0
E   assert 1 == 0
\`\`\`

## Fix

Drop the over-specified \`assert exit_code == 0\`. Keep the three assertions that actually verify what \`fix=True\` is supposed to do:

\`\`\`python
assert \"Re-embedded stale rows\" in stdout
repaired = graph.get_embedding_store_health()
assert repaired[\"mixed_models\"] is False
assert repaired[\"transcript_stale_rows\"] == 0
\`\`\`

Added an inline comment explaining why we don't assert on exit code, so this doesn't get re-added in a future cleanup pass.

## Why not fix the doctor instead?

The doctor's behaviour is correct — it's surfacing a real issue ('no MCP client config has a waggle entry'). The test was wrong to assume \`fix=True\` would clear *every* possible doctor finding. It only clears stale embedding rows.

## Sister test is unaffected

\`test_doctor_diagnoses_mixed_embedding_model_ids\` (line 240, no \`fix=True\`) keeps its \`assert exit_code == 1\` — it's robust because both \"mixed models\" and \"no MCP config\" issues independently force exit 1.

## Files changed

\`\`\`
tests/test_server.py  (+9 / -2)
\`\`\`

## Test plan

- [x] \`ruff check\` passes
- [x] \`ruff format --check\` passes
- [ ] CI test matrix (3.11/3.12/3.13) green — and now actually runs the tests that come after \`test_doctor_fix_reembeds_mixed_embedding_model_ids\` (e.g. the new scope test from #47)

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -310,10 +310,17 @@ def test_doctor_fix_reembeds_mixed_embedding_model_ids(tmp_path: Path, capsys: p
         neo4j_database="",
     )
 
-    exit_code = _run_doctor(config, fix=True)
+    _run_doctor(config, fix=True)
     stdout = capsys.readouterr().out
 
-    assert exit_code == 0
+    # NOTE: we intentionally don't assert on _run_doctor's exit code here.
+    # The doctor reports many orthogonal environment checks (MCP client config
+    # files, embedding model cache, Windows stdout encoding, ...) and exits 1 if
+    # ANY of them surface an issue. On a clean CI runner section [1] always
+    # appends "No MCP client config file contains a 'waggle' server entry",
+    # so the exit code is always 1 regardless of whether the fix path worked.
+    # The assertions below cover what this test actually means to verify:
+    # that fix=True re-embedded the stale rows and the store is now consistent.
     assert "Re-embedded stale rows" in stdout
     repaired = graph.get_embedding_store_health()
     assert repaired["mixed_models"] is False


ard12:
Closing as superseded by #42 — which fixes the same test more rigorously (plants a stub MCP config under monkeypatched HOME/USERPROFILE/APPDATA so the doctor actually returns 0) AND also fixes the Windows test_distribution_cli failure AND a real Codex writer regex bug. I should have checked open PRs before opening this. Sorry for the noise.

PR #73, opened 2026-05-29
ci: fix publish-vscode-extension workflow failing on every push

### LINKED ISSUE #52: ci: publish-vscode-extension workflow fails on every push (startup_failure)
## Symptom

Every push to `Abhigyan-Shekhar/Waggle-mcp` (and downstream forks) creates a failed run of `publish-vscode-extension.yml`, even though the trigger is restricted to `push.tags: v*`.

Sample of recent runs on `main`:

| Time (UTC) | Branch | Conclusion |
|---|---|---|
| 10:42 | main | failure |
| 10:11 | codex/repo-layout-cli-mcp-vscode | failure |
| 09:45 | codex/repo-layout-cli-mcp-vscode | failure |

Each failure has `startup_failure: 1`, `jobs: []`, and an empty `billable` block. No job ever runs.

## Why this matters

- Adds permanent red CI noise on every branch push.
- Masks real workflow failures for contributors who don't know to filter it out.
- The workflow itself never executes, so a real `v*` tag publish would also fail.

## Where to look

`.github/workflows/publish-vscode-extension.yml`, the `Publish to Marketplace` step. The current `if:` condition uses a context that GitHub Actions does not permit in `if:` expressions, which causes the workflow to fail validation at parse-time. When workflow files fail validation, GitHub emits a placeholder failed run on every push regardless of the trigger filter, which explains why branch pushes create runs even though only tag pushes should.

GitHub's documented restriction on `if:` contexts is here: https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

## Acceptance criteria

- [ ] `.github/workflows/publish-vscode-extension.yml` parses cleanly (`python -c "import yaml; yaml.safe_load(open('.github/workflows/publish-vscode-extension.yml'))"`).
- [ ] After the fix lands on `main`, pushing a non-tag branch produces zero new runs of this workflow.
- [ ] The `Publish to Marketplace` step still only publishes when the `VSCE_PAT` secret is set (e.g., a fork build with no PAT should not attempt to publish).
- [ ] A `v*` tag push still triggers the full publish pipeline end-to-end.

## Suggested approach (optional)

Move the secret presence check out of the step-level `if:` and into the step's `run` script. Read `VSCE_PAT` via the `env:` block and exit cleanly when it's empty. A small `concurrency:` block keyed on `github.ref` is a reasonable companion change to avoid racing two simultaneous re-tag pushes.

Good first issue. ~10 lines of YAML.

ard12:
Closes #52. Supersedes the previously-closed #51 (same fix, fresh PR).

## Problem

Every push to `Abhigyan-Shekhar/Waggle-mcp` (and downstream forks) creates a failed run of `publish-vscode-extension.yml`, even though the trigger is restricted to `push.tags: v*`. Sample of affected runs on `main`:

| Time (UTC) | Branch | Conclusion |
|---|---|---|
| 10:42 | main | failure |
| 10:11 | codex/repo-layout-cli-mcp-vscode | failure |
| 09:45 | codex/repo-layout-cli-mcp-vscode | failure |

Each has `startup_failure: 1`, `jobs: []`, empty `billable`. No job ever runs.

## Root cause

```yaml
- name: Publish to Marketplace
  if: ${{ secrets.VSCE_PAT != '' }}
```

The `secrets` context is not allowed in `if:` expressions. GitHub blocks it as a security measure (using `if: secrets.X != ''` would let an attacker probe whether a secret exists through timing/skip patterns). The workflow fails YAML validation at parse-time. For invalid workflow files, GitHub creates a placeholder failed run on every push event, bypassing the trigger filter.

Docs: https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

## Fix

Drop the step-level `if:` and move the secret presence check into the `run:` script:

```yaml
- name: Publish to Marketplace
  env:
    VSCE_PAT: ${{ secrets.VSCE_PAT }}
  run: |
    if [ -z "$VSCE_PAT" ]; then
      echo "::notice::VSCE_PAT not set; skipping marketplace publish."
      exit 0
    fi
    npx @vscode/vsce publish
```

The step always runs but exits cleanly with a notice when no PAT is set (fork builds, PR runs).

## Bonus: concurrency control

Added a small `concurrency` block so two simultaneous `v1.2.3` re-tag pushes serialise the publish step instead of racing the marketplace upload:

```yaml
concurrency:
  group: vscode-publish-${{ github.ref }}
  cancel-in-progress: false
```

`cancel-in-progress: false` is deliberate. An in-flight publish should finish, not get killed mid-upload.

## Files changed

```
.github/workflows/publish-vscode-extension.yml  (+10 / -2)
```

## After this lands

Spurious failed-run accumulation stops on `main` immediately. Existing feature branches keep showing the failure until they're rebased onto post-fix `main`.

## Test plan

- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/publish-vscode-extension.yml'))"` parses cleanly
- [ ] Next branch push to `main` produces zero new runs of this workflow
- [ ] Next `v*` tag push runs the full publish pipeline end-to-end

## Summary by CodeRabbit

* **Chores**
  * Improved publishing workflow reliability by preventing cancellation of in-progress releases and enhancing credential validation checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Abhigyan-Shekhar/Waggle-mcp/pull/73?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/publish-vscode-extension.yml
+++ b/.github/workflows/publish-vscode-extension.yml
@@ -9,6 +9,10 @@ on:
 permissions:
   contents: write
 
+concurrency:
+  group: vscode-publish-${{ github.ref }}
+  cancel-in-progress: false
+
 jobs:
   publish-vscode:
     runs-on: ubuntu-latest
@@ -43,10 +47,14 @@ jobs:
           path: apps/vscode-extension/*.vsix
 
       - name: Publish to Marketplace
-        if: ${{ secrets.VSCE_PAT != '' }}
         env:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
-        run: npx @vscode/vsce publish
+        run: |
+          if [ -z "$VSCE_PAT" ]; then
+            echo "::notice::VSCE_PAT not set; skipping marketplace publish."
+            exit 0
+          fi
+          npx @vscode/vsce publish
 
       - name: Upload VSIX to GitHub release
         if: startsWith(github.ref, 'refs/tags/v')


Abhigyan-Shekhar (approved):
great work @ard12

PR #153, opened 2026-05-30
fix(metrics): escape Prometheus label values

### LINKED ISSUE #115: bug(metrics): label values are not escaped, producing invalid Prometheus output
## Context

`MetricsRegistry._format_labels` builds the Prometheus label block as `key="value"` pairs with a plain f-string (metrics.py:46-51). It does not escape the value. Per the Prometheus exposition format, label values must escape backslash as `\`, double quote as `\"`, and newline as `\n`. Any label value containing one of those characters produces a line a Prometheus scraper rejects.

This is reachable in practice: label values come from `**labels` passed to `increment` / `observe` / `set_gauge` (metrics.py:14-27), and those can carry things like tenant ids, model names, or error strings.

## What to do

Escape label values in `_format_labels` before interpolation: replace `\` with `\`, `"` with `\"`, and newline with `\n`, in that order (backslash first so you don't double-escape the others). Keys are internal and can stay as is, but a comment noting that assumption helps.

## Acceptance criteria

- A label value containing a quote, a backslash, and a newline renders as valid Prometheus output.
- A test asserts the exact escaped string.

## Pointers

- Source: `src/waggle/metrics.py:46-51`, with the entry points at `metrics.py:14-27` and the renderer at `metrics.py:29-44`.
- Prometheus exposition format spec, "Escaping in label values".

## How to claim and report

1. What you changed, in a sentence or two.
4. Anything you skipped or weren't sure about.

vyshali-arc:
## Summary

* Implemented proper escaping for Prometheus label values in `MetricsRegistry._format_labels()`.
* Added escaping for:

  * Backslashes (`\` → `\\`)
  * Double quotes (`"` → `\"`)
  * Newlines (`\n` → `\\n`)
* Added test coverage to verify the exact escaped output.

### Why was it needed?

Prometheus label values must escape backslashes, quotes, and newlines according to the Prometheus exposition format. Without escaping, certain label values could generate invalid Prometheus output that scrapers may reject.

Closes #115

### Commands Run

```bash
pytest tests/test_metrics.py -v
ruff check src/waggle/metrics.py tests/test_metrics.py
ruff format src/waggle/metrics.py tests/test_metrics.py
```

### Pytest Output

```text
======================== test session starts ========================
collected 1 item

tests/test_metrics.py::test_format_labels_escapes_special_characters PASSED

======================== 1 passed ========================
```

### CI Status

All GitHub Actions checks are passing successfully.

## Checklist

* [ ] I updated docs if user-facing behavior changed. (Not needed; no user-facing changes.)

## Files Changed

* `src/waggle/metrics.py`
* `tests/test_metrics.py`


Abhigyan-Shekhar (approved):
Thanks for this fix! This correctly implements Prometheus label value escaping per the exposition format spec.

/lgtm

vyshali-arc:
Thank you for the review and feedback! I appreciate the guidance throughout the process. Glad to contribute to the project, and I'm looking forward to working on more issues.

PR #157, opened 2026-05-30
feat: add deterministic retrieval regression fixtures (#21)

### LINKED ISSUE #21: Add deterministic retrieval regression fixtures
## Problem

Hybrid retrieval combines embeddings, BM25, graph expansion, and recency. Small ranking changes can improve one case while silently regressing another.

## Scope

- Add a compact set of deterministic retrieval regression fixtures.
- Cover at least one lexical match case, one graph-neighborhood case, and one recency-sensitive case.
- Use `WAGGLE_MODEL=deterministic` so the tests are stable offline.
- Keep assertions focused on ordering or required inclusion, not fragile full-score equality.

## Suggested Files

- `tests/test_hybrid_retrieval.py`
- `tests/test_recency.py`
- `src/waggle/retrieval/hybrid.py`
- `src/waggle/retrieval/rankers/temporal.py`

## Acceptance Criteria

- New tests fail if an expected top result disappears from retrieval output.
- The fixtures are small enough to understand without reading benchmark artifacts.
- The tests run without network access or model downloads.
- Existing retrieval tests still pass.

sanrishi:
# Add deterministic retrieval regression fixtures

## Problem

Hybrid retrieval combines embeddings, BM25, graph expansion, and recency. Small ranking changes can improve one case while silently regressing another. There are no regression guards to catch regressions before merge.

## Changes

Three new regression tests that fail if an expected top result disappears:

| Test | File | What it guards |
|---|---|---|
| `test_regression_lexical_match_ranks_expected_first` | `tests/test_hybrid_retrieval.py` | Seeds a record with a unique term ("jade-phoenix"), asserts BM25 lexical layer returns it as top hit |
| `test_regression_graph_neighborhood_includes_connected_nodes` | `tests/test_hybrid_retrieval.py` | Creates nodes linked by `DERIVED_FROM`, asserts graph expansion includes the connected neighbor |
| `test_regression_recency_ranks_recent_transcript_higher` | `tests/test_recency.py` | Seeds two identical transcripts at different timestamps (2025 vs 2026), asserts the recent one outranks the old one |

## Design

- **Deterministic**: All tests use `FakeEmbeddingModel` (16-dim character-modulo hash) or `ConstantEmbeddingModel` (constant output) — same input always produces the same embedding.
- **No network or model downloads**: Tests run offline with zero external dependencies.
- **Ordering assertions, not fragile scores**: Tests check position in ranked output (e.g., "expected record is at position 0") rather than exact float values.
- **Compact**: Each fixture seeds 1–2 records and a handful of nodes. No benchmark artifacts to read.

## Acceptance criteria

- [x] New tests fail if an expected top result disappears from retrieval output
- [x] The fixtures are small enough to understand without reading benchmark artifacts
- [x] The tests run without network access or model downloads
- [x] Existing retrieval tests still pass (all 18 pass)

Closes #21 

## Summary by CodeRabbit

* **Tests**
  * Added regression tests verifying lexical match ranking prioritizes expected results.
  * Added regression tests confirming graph neighborhood expansion includes connected nodes.
  * Added regression tests ensuring recent content ranks higher in retrieval results.

* **Chores**
  * Updated node update handling to accept optional evidence records during updates.

### DIFF
--- a/src/waggle/neo4j_graph.py
+++ b/src/waggle/neo4j_graph.py
@@ -1878,7 +1878,6 @@ def update_node(
     valid_to: datetime | None = None,
     evidence_records: list[EvidenceRecord] | None = None,
 ) -> Node:
-
     if (
         content is None
         and label is None
--- a/tests/test_hybrid_retrieval.py
+++ b/tests/test_hybrid_retrieval.py
@@ -475,9 +475,7 @@ def test_regression_graph_neighborhood_includes_connected_nodes(tmp_path: Path) 
     )
     graph_expansion_layer = debug["layers"].get("graph_expansion", [])
     vector_node_ids = {
-        entry["node_ids"][0]
-        for entry in debug["layers"].get("vector_node", [])
-        if entry.get("node_ids")
+        entry["node_ids"][0] for entry in debug["layers"].get("vector_node", []) if entry.get("node_ids")
     }
     assert graph_expansion_layer, "Graph expansion layer should return results."
     assert node_a.id in vector_node_ids, (
--- a/tests/test_hybrid_retrieval.py
+++ b/tests/test_hybrid_retrieval.py
@@ -380,3 +380,102 @@ def test_score_explanation_includes_recency_contribution():
     }
     assert candidate.score_explanation["recency"] > 0
     assert "final_score" in candidate.score_explanation
+
+
+def test_regression_lexical_match_ranks_expected_first(tmp_path: Path) -> None:
+    graph = make_graph(tmp_path, rerank_enabled=False)
+    with graph._lock, graph._connect() as connection:
+        observed_at = datetime(2026, 1, 1, tzinfo=UTC)
+        graph._store_transcript_record(
+            connection,
+            agent_id="codex",
+            project="alpha",
+            session_id="sess-reg",
+            observed_at=observed_at,
+            turn_index=0,
+            role="user",
+            transcript_text="The deployment uses the jade-phoenix protocol for secure handoff.",
+            turn_pair_id="tp-reg",
+        )
+        graph._store_transcript_record(
+            connection,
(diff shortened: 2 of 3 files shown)


Abhigyan-Shekhar (changes requested):
Changes required

- The new recency regression uses fixed dates in 2025/2026. As time moves forward, both records age and the test becomes less representative of “recent vs old”. Please base the fixtures on `datetime.now(UTC)` minus relative deltas, or inject/freeze `now` where the retriever computes recency.
- The graph-neighborhood test seeds a node relationship but does not assert which seed layer produced the expansion. Please make the setup deterministic enough that the expected seed node is present in `vector_node`, otherwise the graph expansion assertion can pass/fail because of unrelated ranking changes.

Remote CI is green; these are test robustness fixes before merging regression fixtures.

Abhigyan-Shekhar (changes requested):
Critical Issues Found
🔴 #1: Recency Test Will Degrade Over Time

Uses datetime.now(UTC) for both fixtures AND retriever calculation
As time passes, both records age equally
Test becomes less representative of "recent vs old" behavior
Not truly deterministic - behavior changes based on when test runs
Fix: Freeze the reference time used in recency calculations (mock datetime.now() or inject reference time parameter)
🟠 #2: Graph Expansion Test Lacks Determinism Guarantee

Assumes node_a will appear in vector_node seeds
Doesn't guarantee this deterministically
If vector ranking changes, test fails even if graph expansion works correctly
Fix: Use ConstantEmbeddingModel, craft query to match node_a exactly, or seed only the two nodes
Moderate Issues
🟡 Lexical test uses weak substring assertion instead of exact match
🟡 Unrelated whitespace change in neo4j_graph.py

Abhigyan-Shekhar (approved):

PR #159, opened 2026-05-30
fix(rate_limit): clean up inactive rate limit entries

### LINKED ISSUE #117: bug(rate_limit): per-key window and concurrency maps grow without bound
## Context

`RateLimiter` keeps three `defaultdict` maps keyed by request key: `_request_windows`, `_write_windows`, and `_concurrent` (rate_limit.py:23-25). Entries are created on first use and never removed. `check_rate` prunes timestamps inside a key's window but leaves the now-empty `deque` in the map (rate_limit.py:33-37), and `concurrency_slot` decrements the counter to zero but never deletes the key (rate_limit.py:48-49).

For a long-running HTTP server with many distinct keys (per-tenant or per-API-key), these maps grow for the lifetime of the process. It is a slow leak rather than a crash, but it is real.

## What to do

Reclaim entries that are no longer active. Options worth considering:
- After pruning a window in `check_rate`, if the deque is empty, delete the key.
- In `concurrency_slot`'s finally block, if the count hits zero, delete the key.

Be careful with the lock and with `defaultdict` semantics: a read like `len(self._concurrent[key])` will recreate a key you just deleted. Deleting only when empty, under the existing `self._lock`, avoids races. Whatever you choose, make sure an active window is never dropped mid-flight.

## Acceptance criteria

- Empty windows and zero concurrency counts no longer accumulate.
- A test drives a key to empty and asserts it is gone from the relevant map.
- Existing rate-limit behavior is unchanged for active keys.

## Pointers

- Source: `src/waggle/rate_limit.py:23-49`

## How to claim and report

1. What you changed, in a sentence or two.
4. Anything you skipped or weren't sure about, especially around the locking.

vyshali-arc:
### What changed?

* Cleaned up inactive rate limiter entries to prevent unbounded growth of internal maps.
* Removed expired request window entries after pruning old timestamps.
* Removed concurrency keys when their count reaches zero.
* Added test coverage for both request-window cleanup and concurrency cleanup.

### Why was it needed?

The rate limiter stores per-key request windows and concurrency counters. Previously, inactive keys remained in the internal dictionaries even after becoming empty, causing the maps to grow indefinitely in long-running processes. This change reclaims inactive entries while preserving existing rate-limiting behavior.

Closes #117

## Files Changed

* `src/waggle/rate_limit.py`
* `tests/test_rate_limit.py`

### Commands Run

```bash
pytest tests/test_rate_limit.py -v
ruff check src/waggle/rate_limit.py tests/test_rate_limit.py
ruff format --check src/waggle/rate_limit.py tests/test_rate_limit.py
```

### Pytest Output

```text
======================== test session starts ========================
collected 2 items

tests/test_rate_limit.py::test_concurrency_key_removed_when_count_reaches_zero[asyncio] PASSED
tests/test_rate_limit.py::test_expired_request_window_is_cleaned_up[asyncio] PASSED

======================== 2 passed ========================
```

## Notes

* Cleanup logic was implemented under the existing lock to avoid races with the internal `defaultdict`-based maps.
* No additional behavior changes were introduced beyond reclaiming inactive entries.
* Nothing was skipped during implementation.

## Checklist

* [ ] I updated docs if user-facing behavior changed. (Not needed; no user-facing changes.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate limiter memory management by cleaning up expired request-window entries and removing per-key concurrency counters when they reach zero, reducing unnecessary memory growth during sustained usage.
  * Ensured expired entries are compacted so subsequent checks operate on cleaned windows.

* **Tests**
  * Added tests verifying cleanup of expired request windows and removal of per-key concurrency counters to guard against regressions.


Abhigyan-Shekhar (changes requested):
Changes required

- The added tests use `@pytest.mark.anyio`, while the repository's async tests consistently use `@pytest.mark.asyncio`. `pytest-anyio` can parametrize additional backends depending on the environment, which is unnecessary here and can make a simple rate-limiter test suite depend on unsupported backends. Please switch these to `@pytest.mark.asyncio` to match existing `tests/test_rate_limit.py` style.
- Add a regression assertion for the write bucket too. The implementation changes both `_request_windows` and `_write_windows`, but the new expiry cleanup test only covers `is_write=False`.

Remote CI is green and the implementation is small, but the test style/coverage should be tightened before merge.

Abhigyan-Shekhar (dismissed):

Abhigyan-Shekhar:
@vyshali-arc  kindly resolve the conflicts

vyshali-arc:
Hi @Abhigyan-Shekhar,

I've resolved the merge conflicts, updated the branch with the latest changes from `main`, reran the relevant tests and lint checks locally, and pushed the updated branch.

All local checks are passing now. Please let me know if there's anything else you'd like me to address. Thank you for the review!

Abhigyan-Shekhar (approved):

PR #165, opened 2026-05-31
fix(graph): log cosine similarity failures

### LINKED ISSUE #61: bug(graph): _cosine_between_nodes swallows all exceptions and returns None
## Summary

`src/waggle/graph.py` defines an internal helper that computes the cosine similarity between two stored nodes. The body is wrapped in a `try` block whose `except Exception:` returns `None` (line 1675). Any failure — a model dimension mismatch, a corrupted blob, a SQLite I/O error — is collapsed into the same "no answer" sentinel. The caller cannot distinguish "these two nodes have no embedding" from "we hit a real bug while computing similarity".

## Location

- File: `src/waggle/graph.py`
- Lines: 1668–1676

Insert a node with an embedding whose `embedding_dim` differs from another node in the same table. Call the cosine helper on the two IDs. The function silently returns `None`. The caller sees the same return value it gets for any node that genuinely has no embedding stored, so the mismatch never surfaces.

## Expected Behaviour

Real errors propagate (or get logged at WARN with the actual exception) so the operator can tell the difference between "I have no data to compute similarity" and "the data I have is wrong".

## Actual Behaviour

`return None` for any failure mode. No log, no metric.

## Suggested Approach

One way to fix this: split the early-return-on-missing-embedding branch from the catch-all. The "no embedding" case is a legitimate `None`. Everything else should at minimum be `logger.exception(...)` before re-raising or returning `None`, so the failure leaves a trace.

## Acceptance Criteria

- [ ] Missing embedding still returns `None` with no log noise.
- [ ] A model error or DB error logs at WARN or higher with the original exception.
- [ ] A test confirms the WARN log fires when the embedding bytes are deliberately corrupt.

Vijitha14:
## Summary

* Updated `_node_cosine_similarity()` to stop silently swallowing exceptions.
* Added warning-level logging when cosine similarity computation fails.
* Preserved existing non-fatal behavior by returning `None` after logging failures.
* Added a regression test covering embedding decoding failures and verifying that a warning is logged.

## Why was it needed?

Previously, all exceptions inside `_node_cosine_similarity()` were converted to `None` with no visibility into the failure. This made issues such as corrupted embeddings, model errors, or database problems difficult to diagnose. The change preserves the existing behavior for callers while ensuring failures are recorded through warning-level logs.

## Testing

* [x] WAGGLE_MODEL=deterministic pytest -q
* [x] ruff check src/ tests/
* [x] ruff format --check src/ tests/

## Checklist

* [x] I updated docs if user-facing behavior changed.

Closes #61

## Summary by CodeRabbit

* **Bug Fixes**
  * Cosine similarity computation failures now emit warning logs and return None instead of failing silently.
* **Tests**
  * Added a test to verify that failed cosine similarity calculations log a warning and return None.

### DIFF
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 import hashlib
 import heapq
 import json
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -1,5 +1,4 @@
 from __future__ import annotations
-
 import hashlib
 import heapq
 import json
@@ -1811,8 +1810,14 @@ class MemoryGraph:
             vec_a = self.embedding_model.from_bytes(emb_a)
             vec_b = self.embedding_model.from_bytes(emb_b)
             return self.embedding_model.cosine_similarity(vec_a, vec_b)
-        except Exception:
-            return None
+        except Exception as exc:
+            LOGGER.warning(
+                "Failed to compute cosine similarity between nodes %s and %s: %s",
+                a.id,
+                b.id,
+                exc,
+            )
+            raise
 
     def _backfill_transcript_storage(self, connection: sqlite3.Connection, *, batch_size: int = 100) -> None:
         pending_user_pairs: dict[str, str] = {}
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -1818,7 +1818,7 @@ class MemoryGraph:
                 b.id,
                 exc,
             )
-            raise
+            return None
 
     def _backfill_transcript_storage(self, connection: sqlite3.Connection, *, batch_size: int = 100) -> None:
         pending_user_pairs: dict[str, str] = {}
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -381,7 +381,7 @@ def test_node_cosine_similarity_logs_and_reraises_errors(tmp_path, caplog):
     caplog.at_level(logging.WARNING),
     pytest.raises(RuntimeError, match="embedding decode failed"),
 ):
-    graph._node_cosine_similarity(node_a, node_b)
+        graph._node_cosine_similarity(node_a, node_b)
 
     assert any(
         "Failed to compute cosine similarity" in record.message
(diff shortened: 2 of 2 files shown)


Abhigyan-Shekhar (changes requested):
Changes required

- `_node_cosine_similarity()` is used in observation/decomposition edge inference paths. Re-raising from this helper turns a recoverable similarity problem into a failed observation/candidate application. Preserve the existing non-fatal behavior for callers by logging the exception and returning `None`, or update every caller to explicitly catch the new exception and continue.
- The PR description has all required test checkboxes unchecked. Please run and record `WAGGLE_MODEL=deterministic pytest -q`, `ruff check src/ tests/`, and `ruff format --check src/ tests/`.

Local lint/format checks passed on this branch, but the behavior change needs to be made safe before merge.

Vijitha14:
Thanks for the review.

I've updated `_node_cosine_similarity()` to preserve the existing non-fatal behavior by logging failures and returning `None`.

I've also updated the PR description and ran the requested checks locally:

* WAGGLE_MODEL=deterministic pytest -q (454 passed, 4 skipped)
* ruff check src/ tests/
* ruff format --check src/ tests/

Please take another look when you have a chance.

Abhigyan-Shekhar (changes requested):
Status: CHANGES REQUIRED ❌
This PR is almost correct but has a critical flaw that would break existing production behavior.

Critical Issue: Re-raising Breaks Callers
🔴 The Problem:

except Exception as exc:
    LOGGER.warning(...)
    raise  # ❌ This breaks existing callers
The current implementation re-raises the exception after logging, which turns recoverable failures into hard failures. This breaks observation and edge inference paths.

Impact:

Single corrupted embedding → entire observation crashes
Edge inference fails instead of skipping problematic nodes
Production systems experience failures that were previously handled gracefully

Abhigyan-Shekhar (changes requested):
I’m not comfortable approving this as written. `_node_cosine_similarity()` now re-raises decode/runtime failures instead of preserving the previous soft-fail `None` path, and its callers feed that result straight into relationship inference during graph writes. That turns a corrupt embedding from a diagnosable warning into a request-aborting failure path.

Abhigyan-Shekhar (approved):
/lgtm

PR #170, opened 2026-05-31
perf(graph): batch-embed observation nodes in one forward pass

przttttt:
### Changes
Updated `_apply_observation_candidates()` in `graph.py` to compute embeddings with a single `embed_batch()` call before iterating through candidates, then passed the precomputed vectors to `add_node()` via its existing `embedding` parameter. The test suite was updated by adding `embed_batch()` support to `FakeEmbeddingModel` and introducing a new `TestBatchEmbeddingAdoption` class containing four tests.

### Why
Previously, each extracted node triggered an individual `embed()` call, resulting in multiple serial model forward passes for a single observation. Since `embed_batch()` was already available and capable of processing multiple texts in one pass, adopting it reduces embedding overhead and improves throughput while preserving existing behavior.

## Testing

- [x] `WAGGLE_MODEL=deterministic pytest -q` → **489 passed, 4 skipped**
- [x] `ruff check src/ tests/` → **All checks passed**
- [x] `ruff format --check src/ tests/` → **109 files already formatted**

## Checklist

- [x] I linked an issue or explained why one is not needed. *(Batch embedding adoption — sub-issue 1 of 2)*
- [x] I updated docs if user-facing behavior changed. *(No user-facing behavior changed — internal perf path only)*
- [x] I kept the PR focused on one logical change. *(Only `_apply_observation_candidates` + its test file touched)*
- [x] I did not commit secrets, local exports, or generated noise.

## Summary by CodeRabbit

* **Refactor**
  * Improved performance when processing conversations with multiple nodes by batching embedding operations when supported by your embedding model
  * Maintains compatibility with embedding backends that don't support batch operations through graceful fallback to per-node processing

* **Tests**
  * Added comprehensive test coverage for batch embedding functionality and fallback behavior

### DIFF
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -5637,7 +5637,21 @@ class MemoryGraph:
         """
         result = ObservationResult()
         stored_candidate_records: list[tuple[Node, list[str]]] = []
-        for candidate in candidates:
+        _candidate_texts = [str(c["content"]) for c in candidates]
+        _batch_embeddings: np.ndarray | None = None
+        if _candidate_texts:
+            try:
+                _batch_embeddings = self.embedding_model.embed_batch(_candidate_texts)
+                if _batch_embeddings is not None and len(_batch_embeddings) != len(_candidate_texts):
+                    raise ValueError(
+                        f"embed_batch returned {len(_batch_embeddings)} vectors, expected {len(_candidate_texts)}"
+                    )
+            except (AttributeError, NotImplementedError):
+                # embed_batch is not available on this model backend (e.g. a test
+                # stub).  Fall back gracefully: add_node will call embed() itself.
+                _batch_embeddings = None
+
+        for _idx, candidate in enumerate(candidates):
             candidate_tags = list(candidate.get("tags", []))
             speaker_tag = next((tag for tag in candidate_tags if str(tag).startswith("speaker:")), "")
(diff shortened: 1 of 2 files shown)


Abhigyan-Shekhar (changes requested):
Changes required

- Remote CI for the head SHA is not mergeable yet: GitHub Actions reports `CI: completed / action_required` for `d79df4b3295a8d5b8962adfa21c8ef639bac4a05`. Please get the required CI run completed successfully.
- Narrow the `embed_batch()` fallback. The current implementation catches every `Exception`, so real backend/data errors will be silently hidden and retried per-node. Please catch only unsupported-backend cases such as `AttributeError`/`NotImplementedError`, and let unexpected failures surface.
- Add a guard/test for malformed batch output shape or length before indexing `_batch_embeddings[_idx]`, so a backend returning fewer vectors cannot silently fall back per item or write mismatched embeddings.

Local checks I could run before disk pressure interrupted full-suite work: `ruff check`, `ruff format --check`, and deterministic pytest passed for this PR on Python 3.11.

przttttt (commented):
I've addressed all the feedback.

przttttt:
label is wrong* it should be `gssoc26` not `ssoc26`

Abhigyan-Shekhar (approved):
/lgtm

Issue #118, opened 2026-05-30
bug(evidence): build_observation_evidence binds the span to the first match of source_text


ard12:
## Context

`build_observation_evidence` locates the evidence span by calling `transcript.find(source_text)` (evidence.py:61). `str.find` returns the first occurrence. When the same snippet appears more than once in a transcript (a repeated phrase, a quoted earlier turn), the recorded `source_span_start` / `source_span_end` point at the wrong occurrence. Downstream that means evidence highlights or replay offsets can land on the wrong part of the conversation.

## What to do

Let the caller disambiguate which occurrence to record. A reasonable, low-risk approach: add an optional `search_start: int = 0` parameter and call `transcript.find(source_text, search_start)`, so a caller iterating turns can pass the offset where the current turn begins. Keep the existing "not found -> both spans None" behavior (evidence.py:62-66).

If you want to also update callers to pass a sensible offset, note that in the PR, but the core fix is making the correct occurrence reachable.

## Acceptance criteria

- A transcript with a repeated `source_text` can resolve to the intended occurrence.
- The "not found" path still returns `None` for both span fields.
- A test covers the repeated-snippet case and the not-found case.

## Pointers

- Source: `src/waggle/evidence.py:52-75`

## How to claim and report

Comment here before you start so we don't get duplicate PRs. One open claim per person; if a claim goes quiet for about a week we'll free it up.

In your PR description, include:
1. What you changed, in a sentence or two.
2. The files and line ranges you touched.
3. The exact test command you ran and its pasted output. Don't summarize it.
4. Anything you skipped or weren't sure about.

Before you push, run `ruff check .` and `ruff format .`, and make sure `pytest` passes locally. PRs get sent back when there's no real test, when the description doesn't match the diff, or when CI is red.

madsysharma:
Hi @Abhigyan-Shekhar , I would like to work on this as part of GSSoC 2026.

Abhigyan-Shekhar:
@madsysharma  ill assign this issue to you after you've closed issues already assigned to you and no one else claims this

coffeee27:
Hi @Abhigyan-Shekhar,
Please assign this issue to me if it is available.
Thank you!

Abhigyan-Shekhar:
Hey @coffeee27  I suggest you install https://pypi.org/project/waggle-mcp/ before you get started on this issue

PR #176, opened 2026-05-31
fix(evidence): support disambiguating repeated evidence spans

coffeee27:
## Summary

* Added an optional `search_start: int = 0` parameter to `build_observation_evidence`.
* Updated evidence lookup to use `transcript.find(source_text, search_start)` so callers can target the intended occurrence when the same text appears multiple times in a transcript.
* Added tests covering both repeated-snippet resolution and the not-found case.

### Why was it needed?

`str.find()` always returns the first occurrence. When `source_text` appears more than once in a transcript, the recorded evidence span can point to the wrong occurrence. This change makes the intended occurrence reachable while preserving existing behavior when the text is not found.

## Testing

* [x] `python -m pytest tests/test_evidence.py -v`
* [x] `ruff check src/ tests/`
* [x] `ruff format --check src/ tests/`

## Checklist

* [x] I linked an issue or explained why one is not needed. (#118)

## Files Changed

* `src/waggle/evidence.py`
* `tests/test_evidence.py`

## Summary by CodeRabbit

* **New Features**
  * Enhanced transcript searching to locate text snippets starting from a specified position, improving accuracy when handling repeated text occurrences.

* **Tests**
  * Added unit tests for snippet search functionality, covering cases with repeated occurrences and missing text scenarios.

### DIFF
--- a/src/waggle/evidence.py
+++ b/src/waggle/evidence.py
@@ -57,8 +57,9 @@ def build_observation_evidence(
     turn_index: int,
     observed_at: datetime,
     session_id: str = "",
+    search_start: int = 0,
 ) -> EvidenceRecord:
-    start = transcript.find(source_text)
+    start = transcript.find(source_text, search_start)
     if start < 0:
         start = None
         end = None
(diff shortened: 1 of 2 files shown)


Abhigyan-Shekhar (approved):

PR #178, opened 2026-05-31
test(server): added tests for measuring event-loop stall time from synchronous embed calls

### LINKED ISSUE #130: test(server): measure event-loop stall time from synchronous embed calls
## Part of: Async embedding (sub-issue 2 of 2)

Can be done before or after the thread-pool executor sub-issue. If done before, it establishes the baseline stall time. If done after, it validates the fix.

## Context

We claim synchronous embed() blocks the event loop, but we have no measurement. A test that quantifies the stall makes the problem visible and lets us verify the thread-pool fix actually helps.

## What to do

Write a test or benchmark that:
1. Starts the server (or a minimal async loop with the graph/embed wired up).
2. Fires an embed-heavy request (add_node or get_subgraph with a real model, not deterministic).
3. Simultaneously sends a lightweight request (health check or get_stats).
4. Measures how long the lightweight request takes to respond.
5. Reports the stall: how much longer the health check took compared to its expected baseline (<50ms with no load).

If the real model is too slow for CI, use a mock that simulates a 500ms embed call (time.sleep in a patched embed method). The point is measuring event-loop responsiveness, not actual model speed.

## Acceptance criteria

- A test measures and reports the stall time for a lightweight request while embed is running.
- The test works in CI (uses a mock or deterministic model, not a 400MB download).
- If the thread-pool executor sub-issue has landed, the test proves the stall is reduced.

## Compatibility notes

Test-only. No runtime changes. Works regardless of which other performance sub-issues have landed.

## Pointers

- Server: `src/waggle/server.py`
- Embeddings: `src/waggle/embeddings.py:150-173`
- For mocking: patch `EmbeddingModel.embed` with a `time.sleep(0.5)` to simulate slow inference

## How to claim and report

1. What you measured and how.
2. The stall numbers (before and after the executor fix, if applicable).

madsysharma:
# test(server): measure event-loop stall time from synchronous embed calls
 
## Summary

Adds `tests/test_embed_event_loop_stall.py`, which measures how long a lightweight request waits when a synchronous `embed()` runs on the event loop, and shows that offloading the call to a worker thread removes the stall. Test-only; no runtime changes.
 
## What's in this PR
 
- "Lightweight request" modeled by a **heartbeat coroutine** that ticks every 10 ms and records the largest gap between successive ticks. That gap is exactly how long a trivial request (e.g. the `live` health endpoint, which is an `async def` returning JSON) would wait to be serviced. 
- While the heartbeat runs, I drive **one embed-heavy `add_node`** - the real call path: `graph.add_node` -> `_embed_with_metadata` -> `EmbeddingModel.embed` - using two regimes:
 - **blocking**: calls `graph.add_node(...)` directly inside the coroutine, mirroring the current `graph_create_node` ASGI handler in `server.py`.
 - **threaded**: wraps the same call in `asyncio.to_thread(...)`, mirroring the proposed thread-pool-executor fix.
- An **idle** run (no embed, same wall-clock window) captures baseline scheduler jitter for comparison.
 
Per the issue's mocking pointer, the real model is never loaded: I use the deterministic embedder and patch `EmbeddingModel.embed` to `time.sleep(0.5)` before delegating to the original, so the test measures **event-loop responsiveness, not model speed**, and needs no 400 MB download. Each `add_node` uses unique content so the embed LRU cache can't skip the simulated sleep.
 
Three tests assert on a single shared measurement (module-scoped fixture):
- `test_synchronous_embed_stalls_the_event_loop`: blocking stall >= 200 ms and > 5× idle.
- `test_threadpool_offload_keeps_loop_responsive`: threaded stall < 100 ms.
- `test_offload_reduces_stall_versus_blocking`: blocking > 2x threaded and the absolute reduction > 200 ms; prints the report.

Thresholds are deliberately relative/generous (the named no-load baseline is <50 ms; I use a 100 ms ceiling) so the test stays green on noisy CI runners while the nearly 500 ms vs nearly 10 ms gap remains unambiguous.
 
### Stall numbers (before vs after)
 
Since this test contains both the synchronous path (**before**) and the `asyncio.to_thread` path (**after**) side by side, it proves the reduction on its own, regardless of which other performance sub-issues have landed. A representative run:
 
| scenario | event-loop stall |
|---|---|
| idle baseline (no embed) | almost 10 ms |
| synchronous embed (before) | almost 517 ms |
| thread-pool offload (after) | almost 12 ms |
| **reduction** | **almost 42x** |
 
The synchronous path stalls the loop for essentially the full embed duration (nearly 500 ms); the offloaded path keeps it within scheduler jitter of the idle baseline. (exact ms measurements vary per machine; thresholds account for that.)
 
## Testing

```
bash
PS C:\Users\madhu\Documents\Waggle-mcp> python -m pytest tests/test_embed_event_loop_stall.py -v -s
================================================= test session starts =================================================
platform win32 -- Python 3.11.7, pytest-7.4.4, pluggy-1.6.0 -- C:\Program Files\Python311\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\madhu\Documents\Waggle-mcp
configfile: pyproject.toml
plugins: anyio-4.13.0, asyncio-0.23.3, cov-4.1.0
asyncio: mode=Mode.STRICT
collected 3 items

tests/test_embed_event_loop_stall.py::test_synchronous_embed_stalls_the_event_loop PASSED
tests/test_embed_event_loop_stall.py::test_threadpool_offload_keeps_loop_responsive PASSED
tests/test_embed_event_loop_stall.py::test_offload_reduces_stall_versus_blocking
event-loop stall while one embed (~500 ms) runs:
  idle baseline (no embed)      :     19.3 ms
  synchronous embed (blocking)  :    515.2 ms
  thread-pool offload (fixed)   :     18.0 ms
  stall reduction               :    28.6x
PASSED

================================================== 3 passed in 2.84s ==================================================
```

The report prints even without `-s` (it uses `capsys.disabled()`), but `-s` keeps it adjacent to the test it belongs to.
 
`ruff check .` and `ruff format .` were run before pushing; the new file passes both (`All checks passed!` / `1 file already formatted`).
 
## Design decisions
 
- **Why a heartbeat instead of a literal second request**: firing two tasks and timing the second is order-dependent and flaky, since a synchronous coroutine step runs to completion before the loop can schedule anything else. A heartbeat watchdog is the standard, robust way to observe loop responsiveness, and maps directly to "how much longer the lightweight request took vs baseline."
- **Real call path, not a bare `embed`**: routing through `graph.add_node` keeps the test faithful to the `graph_create_node` handler. `add_node` runs cleanly inside `asyncio.to_thread` (it opens its own sqlite connection per call), which I verified.
- **MCP `call_tool` path already non-blocking** (uses `anyio.to_thread.run_sync`): the blocking path this test targets is the ASGI HTTP handler `graph_create_node`, which calls `graph.add_node` directly on the loop: that's the one the executor fix should address.
- **Did NOT** add a fourth scenario that runs against the full Starlette app / real `WaggleServer`; the auth/scope plumbing adds brittleness for no extra signal over the minimal-loop harness the issue explicitly allows.

## Summary by CodeRabbit

* **Tests**
  * Added test suite to measure event loop responsiveness during embedding operations, comparing different execution approaches for performance impact.

### DIFF
--- a/tests/test_embed_event_loop_stall.py
+++ b/tests/test_embed_event_loop_stall.py
@@ -1,13 +1,13 @@
 """Measure event-loop stall caused by synchronous ``embed()`` calls.
 
-Part of the *async embedding* work (sub-issue 2 of 2). 
+Part of the *async embedding* work (sub-issue 2 of 2).
 
 We claim that calling ``EmbeddingModel.embed`` synchronously inside an async request handler blocks the event loop, starving lightweight requests (health checks, ``get_stats``).
 This module quantifies that stall so the problem is visible and so the thread-pool-executor fix can be shown to help.
 
 How it works
 ------------
-A "heartbeat" coroutine ticks every ``HEARTBEAT_INTERVAL`` seconds and records the largest gap between successive ticks. 
+A "heartbeat" coroutine ticks every ``HEARTBEAT_INTERVAL`` seconds and records the largest gap between successive ticks.
 Each tick stands in for a lightweight request: the gap is how long such a request would have waited to be serviced.
 While the heartbeat runs we drive one embed-heavy ``add_node`` through two code paths:
 
--- a/tests/test_embed_event_loop_stall.py
+++ b/tests/test_embed_event_loop_stall.py
@@ -1,0 +1,183 @@
+"""Measure event-loop stall caused by synchronous ``embed()`` calls.
+
+Part of the *async embedding* work (sub-issue 2 of 2). 
+
+We claim that calling ``EmbeddingModel.embed`` synchronously inside an async request handler blocks the event loop, starving lightweight requests (health checks, ``get_stats``).
+This module quantifies that stall so the problem is visible and so the thread-pool-executor fix can be shown to help.
+
+How it works
+------------
+A "heartbeat" coroutine ticks every ``HEARTBEAT_INTERVAL`` seconds and records the largest gap between successive ticks. 
+Each tick stands in for a lightweight request: the gap is how long such a request would have waited to be serviced.
+While the heartbeat runs we drive one embed-heavy ``add_node`` through two code paths:
+
+- ``blocking``: mirrors the current ``graph_create_node`` ASGI handler, which calls ``graph.add_node(...)`` (and therefore ``embed``) directly on the loop.
+- ``threaded``: mirrors the proposed fix, wrapping that same call in ``asyncio.to_thread`` so the blocking work runs off the loop.
+
+The real model is never loaded. We use the deterministic embedder and patch ``EmbeddingModel.embed`` to ``time.sleep`` for a fixed interval, so the test
+measures event-loop responsiveness (not model speed) and runs anywhere with no 400 MB download.
+
+Run with ``-s`` to see the reported numbers:
+
+    pytest tests/test_embed_event_loop_stall.py -v -s
+"""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from collections.abc import Callable
+from itertools import count
+
+import pytest
+
+from waggle.embeddings import EmbeddingModel
+from waggle.graph import MemoryGraph
+from waggle.models import NodeType
+
+# ---------------------------------------------------------------------------
+# Tunables
+# ---------------------------------------------------------------------------
+
+# Simulated cost of one synchronous embed forward pass.
+EMBED_SLEEP_SECONDS = 0.5
+# Heartbeat cadence; also the expected baseline latency of a lightweight request.
+HEARTBEAT_INTERVAL = 0.01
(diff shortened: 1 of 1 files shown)


madsysharma (commented):
Hi @Abhigyan-Shekhar , please review this PR. Thank you.

Abhigyan-Shekhar (approved):

madsysharma:
Hi @Abhigyan-Shekhar , have carried out the installation without issue. Please review this updated PR, and if it's good to go, please add the "gssoc:approved" label and merge it. Thank you.

Amlan-git:
Looks good to me.  ruff is clean, and the coverage is solid across all four functions. Adding `gssoc:approved` and merging. Thanks @madsysharma!

PR #181, opened 2026-06-01
fix(graph): raise RuntimeError on read→write lock upgrade to prevent …

### LINKED ISSUE #67: bug(graph): _ReadWriteLock allows reader-then-writer self-deadlock on the same thread
## Summary

The reader-writer lock introduced in PR #12 (`src/waggle/graph.py` around lines 687–757) handles "writer holds, asks for read" correctly via a `_is_writer` shortcut. The opposite direction, "reader holds, asks for write", is not handled. A thread that has acquired the read context and then tries to acquire the write lock blocks forever, waiting for `_readers` to drop to zero, but the only reader it is waiting on is itself.

Today no caller exercises `lock.read()` (the PR description notes this), so the deadlock is unreachable. The API is exposed and the deadlock is silent (no exception, no log), so the first contributor who migrates a read-only call site to `lock.read()` and then later upgrades to a write will hit this without diagnostic.

## Location

- File: `src/waggle/graph.py`
- Class: `_ReadWriteLock`
- Lines: 687–757

```python
lock = _ReadWriteLock()

def workload():
    with lock.read():
        with lock:   # write — deadlocks forever
            ...

threading.Thread(target=workload).start()
```

The thread hangs.

## Expected Behaviour

Either the second acquire raises a clear `RuntimeError("cannot upgrade read lock to write")`, or the lock detects the same-thread upgrade and accepts it as an exclusive transition.

## Actual Behaviour

Silent deadlock.

## Suggested Approach

One way to fix this: track read ownership by `threading.get_ident()` (parallel to the `_write_owner` mechanism). On a write acquire, if the only outstanding readers are the current thread, allow the upgrade and stash the prior read depth so the eventual write release restores it.

A simpler defensive option: detect the case and raise `RuntimeError` with a clear message. Refuse the lock-upgrade. Cheap and obvious.

## Acceptance Criteria

- [ ] A test attempts the read-then-write pattern from a single thread and either passes (upgrade allowed) or raises `RuntimeError` immediately (refused).
- [ ] No thread hangs forever in the test.
- [ ] The behaviour chosen (upgrade vs refuse) is documented as a docstring on `_ReadWriteLock`.

riya-dumbare:
Fixes #67

### What changed?
- Added `_reader_threads: set[int]` to `_ReadWriteLock.__init__` to track which threads currently hold a read lock.
- `_ReadContext.__enter__` now registers the calling thread ID into `_reader_threads` on read acquire.
- `_ReadContext.__exit__` discards the thread ID from `_reader_threads` on read release.
- `_ReadWriteLock.__enter__` (write path) now checks if the calling thread is already in `_reader_threads` and raises `RuntimeError` immediately instead of waiting forever.
- Updated the `_ReadWriteLock` docstring to document that read→write lock upgrade is explicitly refused.
- Added `tests/test_rw_lock.py` with 18 tests covering the bug fix and regression cases.

### Why was it needed?
A thread holding a read lock that then attempted to acquire the write lock would block forever — it was waiting for `_readers` to drop to zero, but it was the only reader. The deadlock was silent (no exception, no log), making it very hard to diagnose. This fix detects the upgrade attempt and raises a clear `RuntimeError` immediately.

## Testing
- [x] `WAGGLE_MODEL=deterministic pytest -q` — 18/18 new tests pass, 447 existing tests pass (10 pre-existing Windows-only failures unrelated to this change)
- [x ] `ruff check src/ tests/`
- [x ] `ruff format --check src/ tests/`

## Checklist
- [x] I linked an issue or explained why one is not needed. (Fixes #67)
- [x] I updated docs if user-facing behavior changed. (docstring updated on `_ReadWriteLock`)
- [x] I kept the PR focused on one logical change.
- [x] I did not comm

## Summary by CodeRabbit

* **Refactor**
  * Enhanced reader/writer lock implementation with improved concurrency handling, including recursive re-entrancy support and better thread safety validation.

* **Tests**
  * Added comprehensive test suite covering lock concurrency scenarios, edge cases, and thread safety validations.
  * Added regression tests for lock upgrade scenarios.


Amlan-git (commented):
Hey @riya-dumbare, the lock implementation looks solid and the fix for #67 is exactly right. 
But, please address the two test issues CodeRabbit flagged:

1. **`test_write_exclusivity` (line ~48-54)** — the `time.sleep(0.01)` to sequence threads A and B is a timing race that can flake under CI load. Please use a `threading.Event` to signal when thread A is actually inside the lock before starting B.

2. **`test_per_thread_reader_depth_is_independent` (line ~324-328)** — thread A is fully joined before B starts, so they never overlap. This doesn't actually verify per-thread independence. Please restructure so B holds a read lock concurrently while A's upgrade attempt is being refused.

CodeRabbit's comment on the PR has suggested code for both fixes. Thanks!

Amlan-git (changes requested):
Thanks for the focused fix — #67's plain `read -> write` case is handled correctly, and the `not t.is_alive()` assertion shape is right. 
Two issues to address before merge:

**1. `_acquire_read()` still self-deadlocks on `read -> queued writer -> nested read`.** If thread A holds a read, thread B queues for write, and A then re-enters `lock.read()`, A blocks on `_waiting_writers > 0` while B waits for A's original read to drain. I reproduced this locally on PR head. Fast-path `self._reader_threads.get(tid, 0) > 0` before the writer-preference wait.

The intended test for this in `tests/test_rw_lock.py:312-353` is currently dead code: `thread_c`/`thread_d` are defined but never started, and `a_inside.wait(timeout=2)` never observes a `.set()`. Please add a real regression once the fast-path is in.

**2. `ScoredNodeView` regressed from `@dataclass(slots=True)` to a Pydantic `BaseModel`.** The original comment was explicit that avoiding Pydantic was the point: reducing allocations in `_sort_scored_nodes` at larger candidate counts. Please restore the slots dataclass shape; keeping it in `models.py` is fine, and a module-level import in `graph.py` is preferable if it does not reintroduce an import cycle.

Also, CI is red on `Lint (ruff)`: `tests/test_rw_lock.py` needs formatting, so running `ruff format tests/test_rw_lock.py` should clear it.

The `RuntimeError`-on-upgrade design and `dict[int, int]` reader-depth tracking are the right calls.

Amlan-git:
One note for next time — totally fixable, and very common on a first busy-repo PR: this branch picked up a lot of unrelated commits while catching up with `main`, which made review and conflict resolution harder than it needed to be. The likely culprit was rebasing/merging while the branch was several days behind and then resolving overlapping `graph.py` changes; that kind of conflict can silently drop earlier local fixes, which seems to be what happened around `ScoredNodeView`.

Force-pushing after a rebase is expected; as a habit, prefer `git push --force-with-lease` — same effect, with a safety check against clobbering someone else's push.

If a branch gets tangled on a future PR, the cleanest recovery is to start a fresh branch off `main` and cherry-pick your own commits onto it. `git log --author="$(git config user.email)" --oneline` can help identify them, then force-push to replace the messy branch:

> ```
> git fetch origin
> git checkout -b yourbranch-v2 origin/main
> git cherry-pick <your-commit-shas>
> git push --force-with-lease origin HEAD:yourbranch
> ```

Also, worth a quick check before any push in such cases: `git log --oneline origin/main..HEAD` should list only your PR's commits. Anything else means the branch has absorbed work that belongs in someone else's PR.

Nothing to do about it on this PR — squash-merge will clean up the history. Just a useful workflow note for next time.

Thanks @riya-dumbare !

Amlan-git (changes requested):
Thanks for the updates. The core lock behavior looks correct now: read-to-write upgrades raise immediately, re-entrant reads bypass a queued writer, `ScoredNodeView` is back to a slots dataclass, and CI is green.

One remaining item before approval: please address CodeRabbit’s latest test comment. Since this PR is specifically about preventing lock deadlocks, the regression tests need to fail deterministically if a worker thread asserts, crashes, or hangs. Please make the test helper use non-daemon threads by default, assert timed joins with `not t.is_alive()`, and propagate worker-thread exceptions back to the main test thread, especially in `test_reentrant_read_bypasses_waiting_writer`.

The older CodeRabbit comments about the re-entrant read fast-path and `ScoredNodeView` are handled. The induced-subgraph comment appears unrelated to the current PR diff and should not block this lock fix.

Happy to re-review once the test-thread cleanup is in.

ard12 (dismissed):
This is the kind of fix the lock protocol needed. The bug — read-holder attempts write acquire and deadlocks silently — is exactly the failure mode #67 describes, and tracking `_reader_threads: set[int]` plus the explicit `RuntimeError` on detection is the right design. The 380-line test suite covers the regression cases and the new thread-tracking semantics properly.

`graph.py` is a high-blast-radius file, so I checked carefully:

- Thread tracking is correctly added on read acquire (`_ReadContext.__enter__`) and removed on release (`__exit__`).
- The write path checks `threading.get_ident() in self._reader_threads` before blocking on `_readers`, which is what avoids the deadlock.
- The new `ScoredNodeView` dataclass in `models.py` and the `Node.model_rebuild()` call are needed to support the graph.py changes — not unrelated scope, just a slightly broader file footprint than the PR title implies.

CI all green across all four platforms. Approved.

Once this lands, the `clear_scope_dry_run_and_audit_trail` test pattern at the bottom of test_graph.py would be a nice template for future invariant tests.

Code. Commit. Pray. Repeat.

ard12:
@Amlan-git the test-thread cleanup commits are in (`test(rw_lock): propagate worker-thread exceptions to main test thread` and the follow-up ruff format pass). CI is green across all four platforms. Whenever you have a moment, could you take another look?

May your CI/CD pipelines flow like water 🌊

Amlan-git (commented):
Thanks for the follow-up. The production lock behavior looks right now: read-to-write upgrades raise immediately, re-entrant reads bypass queued writers, write re-entrancy is preserved, and `ScoredNodeView` is back to a slots dataclass.

One remaining test-harness cleanup before merge: the new `_join()` helper is only used in the final test. Since this PR is specifically about preventing lock hangs, the whole concurrency suite should fail deterministically when a worker thread crashes, asserts, or fails to exit.

Could you please make `_start()` non-daemon by default and route the remaining worker joins through `_join()` or equivalent `not t.is_alive()` + exception propagation checks?

Small cleanup while touching that file: assert `a_inside.wait(timeout=2)` before starting thread B in the exclusivity test, and drop the unused `_orig_worker`.

The core fix for #67 looks good; this is just making the regression tests as reliable as the lock fix. After that, this would be good to go. 

Good work, thanks @riya-dumbare

Amlan-git (commented):
Thanks for the follow-up. The production lock behavior looks good now: read-to-write upgrades raise immediately, re-entrant reads bypass queued writers, write re-entrancy is preserved, and ScoredNodeView is back to a slots dataclass.

The latest test-harness cleanup also addresses almost all of my previous review: _start() is non-daemon by default in tests/test_rw_lock.py:29, _join() now asserts timed exits and propagates worker-thread exceptions in tests/test_rw_lock.py:45-51, the exclusivity test waits for thread A to enter the lock before starting thread B in tests/test_rw_lock.py:66-77, and the unused _orig_worker is gone.

One small remaining cleanup from my earlier comment is still missing: in tests/test_rw_lock.py:271-272, test_no_thread_hangs starts the worker with _start(workload), but does not save the returned thread object. Because of that, the test waits for done.set(), but it never joins the worker afterward.

For these lock tests, joining matters because _join() does two useful things: it checks that the thread really finished, and it re-raises any exception that happened inside the worker thread so the test fails clearly. This was the intent behind my earlier request to route the remaining worker joins through _join() or an equivalent not t.is_alive() plus exception-propagation check.

Could you please update that test to follow the same pattern as the others: 

- Change tests/test_rw_lock.py:271 to save the thread:   t = _start(workload)
- Keep the existing   assert done.wait(timeout=2)    on tests/test_rw_lock.py:272
- Add _join(t) after that assertion

After that, this is good to go from my side. CI is green and the core fix looks solid.

Abhigyan-Shekhar (approved):
/lgtm

Amlan-git (approved):
Thanks for the follow-up. Re-reviewed latest head. The remaining test-harness cleanup is addressed: test_no_thread_hangs now keeps the worker thread and routes it through _join(t), so it gets the same bounded join and exception propagation as the rest of the concurrency suite.

The lock behavior looks correct: read-to-write upgrades fail fast, re-entrant reads bypass queued writers, write re-entrancy is preserved, and ScoredNodeView remains a slots dataclass.

Verified locally with the focused suite & the latest CI is green.
Looks good to me. 
Approving

Abhigyan-Shekhar (approved):

Abhigyan-Shekhar (approved):

Abhigyan-Shekhar (approved):

Abhigyan-Shekhar:
@ard12  can you merge it

Issue #108, opened 2026-05-30
test(errors): cover code and status_code for every WaggleError subclass


ard12:
## Context

`src/waggle/errors.py` defines a base `WaggleError` and nine subclasses (`AuthenticationError`, `AuthorizationError`, `RateLimitExceededError`, `PayloadTooLargeError`, `ServiceUnavailableError`, `ValidationFailure`, `ConflictResolutionError`, `DanglingEdgeError`, `HashVerificationError`, `SchemaVersionError`). Each sets a specific `code` string and HTTP `status_code` (errors.py:27-74). Nothing in the test suite asserts those values, so a typo like a 401 silently becoming a 400 would ship unnoticed.

## What to do

Add a test (a parametrized one reads best) that constructs each subclass and asserts its `.code` and `.status_code`. For example, `AuthenticationError().status_code == 401` and `.code == "authentication_failed"`. Cover all of them.

## Acceptance criteria

- A new test in `tests/` exercises every `WaggleError` subclass.
- Each assertion checks both `code` and `status_code`.
- The test fails if someone changes a status code by accident (try it: flip one and confirm red).

## Pointers

- Source: `src/waggle/errors.py:27-74`
- Look at existing tests in `tests/` for the fixture and style conventions.

## How to claim and report

Comment here before you start so we don't get duplicate PRs. One open claim per person; if a claim goes quiet for about a week we'll free it up.

In your PR description, include:
1. What you changed, in a sentence or two.
2. The files and line ranges you touched.
3. The exact test command you ran and its pasted output. Don't summarize it.
4. Anything you skipped or weren't sure about.

Before you push, run `ruff check .` and `ruff format .`, and make sure `pytest` passes locally. PRs get sent back when there's no real test, when the description doesn't match the diff, or when CI is red.

Vijitha14:
Hi! I'd like to work on this issue. The scope and acceptance criteria are clear, and I'd be happy to add the test coverage for all WaggleError subclasses. Could you please assign it to me?

ard12:
@Vijitha14 Thanks for your interest in this issue and for the clear scope.

I’m assigning this task to you. Please go ahead with adding test coverage for all WaggleError subclasses, ensuring each case is properly validated and existing behavior remains unchanged.

May your exceptions be well-bounded and your test suite always green.

Elshayib:
[...]

PR #188, opened 2026-06-01
test(errors): cover code and status_code for WaggleError subclasses

Vijitha14:
### What changed?

* Added a new parametrized test covering every `WaggleError` subclass defined in `src/waggle/errors.py`.
* The test verifies that each subclass exposes the expected `.code` and `.status_code` values.

### Why was it needed?

* The project did not have test coverage validating the error codes and HTTP status codes for `WaggleError` subclasses.
* This test prevents accidental regressions, such as changing a status code or error code string without detection.

## Testing

* [x] `python -m pytest tests/test_errors.py -q`
* [x] `python -m ruff check tests/test_errors.py`
* [x] `python -m ruff format --check tests/test_errors.py`

## Checklist

* [x] Linked to Issue #108.
* [x] I updated docs if user-facing behavior changed. (Not needed; no user-facing behavior changed.)

## Summary by CodeRabbit

* **Tests**
  * Added parametrized tests that validate error classes' public codes and HTTP status mappings, and assert the set of exported error types matches expectations.

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -31,12 +31,10 @@ EXPECTED_ERROR_METADATA = {
 
 assert set(WaggleError.__subclasses__()) == set(EXPECTED_ERROR_METADATA)
 
+
 @pytest.mark.parametrize(
     ("error_cls", "expected_code", "expected_status_code"),
-    [
-        (cls, code, status)
-        for cls, (code, status) in EXPECTED_ERROR_METADATA.items()
-    ],
+    [(cls, code, status) for cls, (code, status) in EXPECTED_ERROR_METADATA.items()],
 )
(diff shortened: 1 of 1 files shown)


ard12 (approved):
Clean test design. The parametrized approach covers all 10 WaggleError subclasses in one test, and the module-level set comparison catches any future subclass that gets added without a matching test entry.

One thing to consider for a follow-up: that module-level `assert` runs at import time, not as a pytest test. If a new subclass is added, the entire test file fails to import rather than showing a clear test failure. Moving it into its own test function would give a better error message. Not blocking.

CI all green. Approved.

No bugs were harmed in the making of this PR (hopefully) 😄

PR #202, opened 2026-06-01
fix(devops): reorder Dockerfile COPY steps to preserve dependency layer caching

### LINKED ISSUE #201: devops(docker): dependency layers are invalidated on every source change
Dockerfile:8-32

The builder stage copies src/ before pip install ".[neo4j]", so every code change busts the layer cache that resolves and installs dependencies. The final stage repeats the pattern with pip install --no-deps -e ., which also weakens rebuild caching during normal development.

Impact: image rebuilds pay the full dependency-install cost far more often than necessary, which slows local iteration and CI image publishing. Reordering the COPY and install steps would make the Docker build much more cache-friendly without changing runtime behavior.

sanrishi:
Closes #201

Reorder the builder stage in Dockerfile so that \pip install\ runs before \COPY src ./src\. This ensures source code changes don't bust the dependency layer cache, dramatically speeding up rebuilds.

Also split the model pre-download into its own RUN step so the model cache layer is independently cached.

## Summary by CodeRabbit

* **Chores**
  * Optimized Docker build process with improved layer caching for faster builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,16 +10,20 @@ FROM python:3.11-slim AS builder
 ENV PIP_NO_CACHE_DIR=1
 
 WORKDIR /build
-COPY pyproject.toml README.md LICENSE ./
-COPY src ./src
 
-# 1) CPU-only PyTorch (avoids pulling the 2 GB CUDA wheels)
-# 2) Install project + neo4j extras
-# 3) Pre-download the embedding model so it is baked into the image layer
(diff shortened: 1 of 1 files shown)


ard12 (approved):
Standard Docker best practice: copy dependency metadata first, install deps, then copy source. Source-only changes no longer bust the pip install layer. Splitting the model download into its own RUN step is a nice touch too since that layer is the slowest to rebuild.

CI all green. Approved.

Ship it, but carefully 🚢

PR #203, opened 2026-06-01
perf(graph): add composite indexes for transcript project and agent queries

### LINKED ISSUE #200: perf(graph): transcript browsing misses the indexes it needs for project and agent filters
src/waggle/graph.py:7408-7471, 491-495

list_transcript_records() and count_transcript_records() filter on tenant_id plus project or agent_id, but the schema only adds an index for tenant_id, session_id, and turn_index. Common transcript-browsing paths therefore fall back to scans once history grows beyond cache-friendly size.

Impact: the transcript UI and API will slow down as the conversation archive grows, even though the code already knows which composite predicates it should optimize. Adding the missing composite index pair is a contained schema fix.

sanrishi:
Closes #200

Adds \idx_transcripts_tenant_project\ and \idx_transcripts_tenant_agent\ to the SQLite schema so \list_transcript_records\ and \count_transcript_records\ can seek via index when filtering by project or agent_id alone. Adds matching indexes to the Neo4j backend.


ard12 (approved):
The two new composite indexes (tenant_id + project, tenant_id + agent_id) cover the filter patterns in list_transcript_records and count_transcript_records. Adding them to both SQLite and Neo4j keeps the backends consistent.

These are CREATE INDEX IF NOT EXISTS, so they're safe on existing databases. No migration needed.

CI all green. Approved.

Branch wisely, merge safely 🌿

PR #205, opened 2026-06-01
perf(graph): cap replay query rows to avoid scanning entire tenant

### LINKED ISSUE #198: perf(graph): replay and graph queries scan whole tenant tables on every request
src/waggle/graph.py:3554-3858

_query_graph_only(), _query_replay_hits(), and _query_replay_session_scores() all fetch broad tenant-level row sets and then recompute embeddings or lexical scores in Python row by row. Even when project, agent_id, or session_id are supplied, the code still starts from a large in-memory scan.

Impact: query latency scales with total tenant history instead of the requested scope. In HTTP mode this can turn ordinary retrieval requests into event-loop stalls.

sanrishi:
Closes #198

Add \LIMIT\ clauses to \_query_replay_hits\ and \_query_replay_session_scores\ so they stop fetching and scoring every matching transcript record in the tenant. Combined with the composite indexes added in #200, this ensures query latency scales with the requested scope rather than total tenant history.

### DIFF
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -3781,14 +3781,19 @@ class MemoryGraph:
             elif agent_id.strip():
                 filters.append("agent_id = ?")
                 params.append(agent_id.strip())
+            # Cap the rows fetched so we don't scan and score the entire tenant.
+            # The downstream scorer uses up to max_hits results; read a generous
+            # multiple so semantic ranking still has a good pool to draw from.
+            fetch_limit = min(max(500, max_hits * 4), 5000)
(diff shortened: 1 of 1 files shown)


ard12 (approved):
Good catch. Without a LIMIT, both replay queries would scan the entire tenant's transcript table before scoring. The cap formula `min(max(500, max_hits * 4), 5000)` gives the scorer a generous candidate pool without unbounded reads. The hard 5000 cap on session scores is reasonable for the same reason.

CI all green. Approved.

May your builds be faster than your coffee break ☕⚡

PR #207, opened 2026-06-01
perf(graph): fix edge expansion to find edges with either endpoint in seed set

### LINKED ISSUE #196: perf(graph): edge expansion only sees edges whose both endpoints are already selected
src/waggle/graph.py:8361-8377, 3554-3740

_fetch_edges_for_nodes() filters with source_id IN (...) AND target_id IN (...), which only returns edges fully contained inside the current seed set. The graph walk in _query_graph_only() therefore cannot traverse outward through normal edges, so max_depth > 0 expansion misses most neighbors.

Impact: graph queries return incomplete subgraphs and users retry with larger windows or deeper searches to recover missing context. The bug also wastes work because the code performs expansion logic that cannot actually cross the seed boundary.

sanrishi:
Closes #196

Change \_fetch_edges_for_nodes\ from \source_id IN (...) AND target_id IN (...)\ to \source_id IN (...) OR target_id IN (...)\ so the graph walk can traverse outward through edges that connect a seed node to a non-seed neighbor. Previously the expansion was effectively dead code because it could never cross the seed boundary.

### DIFF
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -8366,13 +8366,14 @@ class MemoryGraph:
         if not node_ids:
             return []
         placeholders = ", ".join("?" for _ in node_ids)
+        # Use OR so the graph walk can traverse outward: return any edge
+        # whose source OR target is in the current seed set.
         rows = connection.execute(
             f"""
             SELECT id, source_id, target_id, relationship, weight, metadata, created_at, tenant_id
             FROM edges
             WHERE tenant_id = ?
-              AND source_id IN ({placeholders})
-              AND target_id IN ({placeholders})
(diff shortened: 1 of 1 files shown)


ard12 (approved):
This is a correctness fix, not just a perf change. The old query used AND between source_id and target_id, which only returned edges where both endpoints were already in the seed set. That defeats the purpose of graph expansion since you'd never discover edges connecting a known node to an unknown one.

The OR fix is correct: return any edge where at least one endpoint is in the seed set, then let the caller traverse outward.

CI is all green. Approved.

Happy coding, may your bugs be few 🐛✨

PR #209, opened 2026-06-01
security(export): expand secret scanning to all export paths and data sources

### LINKED ISSUE #194: security(export): secret scanning only protects transcript text and only for ABHI commits
src/waggle/server.py:312-385, 2099-2145, 3542-3578, 4945-4964

_scan_export_transcripts_for_secrets() only inspects transcript rows, so secrets stored in node content, source prompts, metadata, or markdown vault files are never scanned. On top of that, the check is only invoked for the plain ABHI commit path; bundle, markdown, and JSON export paths still write full graph data without any secret gate.

Impact: API keys and credentials can be exfiltrated through alternate export formats even when the ABHI export path would have refused them. The current protection is easy to bypass by choosing a different export surface.

sanrishi:
Closes #194

Replace \_scan_export_transcripts_for_secrets\ with \_scan_export_for_secrets\ that inspects node labels, content, source_prompt, and serialised metadata in addition to transcript text.

Apply \_assert_export_safe\ to:
- ABHI exports (existing — unchanged)
- Bundle and markdown-vault exports (was missing)
- HTTP \graph_export\ for both abhi and json formats (was missing)
- CLI \export-context-bundle\ and \export-markdown-vault\ commands (was missing)

### DIFF
--- a/src/waggle/server.py
+++ b/src/waggle/server.py
@@ -361,12 +361,14 @@ def _scan_export_for_secrets(
             if not field_text.strip():
                 continue
             for label, secret, preview in _scan_text_for_secrets(field_text):
-                findings.append({
-                    "pattern": label,
-                    "source": f"node.{field_name}",
-                    "node_id": str(node.get("id", "")),
-                    "preview": preview[:180],
-                })
+                findings.append(
+                    {
+                        "pattern": label,
+                        "source": f"node.{field_name}",
+                        "node_id": str(node.get("id", "")),
+                        "preview": preview[:180],
+                    }
+                )
                 if len(findings) >= max_findings:
                     return findings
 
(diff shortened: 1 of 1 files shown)


ard12 (changes requested):
The expansion is the right direction: scanning node label, content, source_prompt, and serialized metadata in addition to transcripts catches secrets that previously slipped through if they were stored as graph nodes. The helper extraction (`_scan_text_for_secrets`) is also a clean refactor.

But CI is fully red (lint + all four test platforms). Likely causes:

1. The function was renamed from `_scan_export_transcripts_for_secrets` to `_scan_export_for_secrets`. Existing callers and tests still reference the old name. Grep for the old name and update.
2. Lint failure is probably the dict-literal style (uses `{...}` directly rather than line-broken). Run `ruff format src/waggle/server.py` locally.

Once those two are sorted CI should go green. The logic itself reads correct.

May your merge conflicts be ever in your favor 😄

sanrishi:
@ard12 @Abhigyan-Shekhar Applied ruff formatting and fixed the transcript scan limit (was defaulting to 200 rows, now passes 100000). Ready for re-review!

ard12 (changes requested):
The follow-up commits (ruff format + transcript limit fix) didn't resolve CI. Lint still fails and all four test platforms are red.

Quick local loop before pushing:

```bash
ruff check src/ tests/
ruff format --check src/ tests/
WAGGLE_MODEL=deterministic pytest -q
```

The lint failure should give you the file and line immediately. For the test failures, my earlier guess about the function rename (`_scan_export_transcripts_for_secrets` → `_scan_export_for_secrets`) is still worth checking. Any test or caller still referencing the old name will fail.

Re-review queued once CI is green.

Abhigyan-Shekhar (commented):
Maintainer pass on June 11, 2026: not mergeable in this pass because the current GitHub CI run is failing. Please update the branch so the failing jobs rerun green, then re-request review.

Abhigyan-Shekhar:

Abhigyan-Shekhar:

PR #210, opened 2026-06-01
security(http): cap request body size before parsing JSON or writing temp files

### LINKED ISSUE #193: security(http): cap request body size before parsing JSON or writing temp files
src/waggle/server.py:3075-3693

The HTTP route handlers call await request.json() and write payloads to temporary files, but they never apply the max_payload_bytes guard that the MCP transport uses. A client can send arbitrarily large bodies to import, diff, query, or preview endpoints.

Impact: the service can be forced into high memory usage, large temp-file writes, and long blocking parses. That is a trivial unauthenticated DoS on the public HTTP surface.

sanrishi:
Closes #193

Add \_RequestBodySizeMiddleware\ ASGI middleware that reads the \Content-Length\ header and rejects requests exceeding \max_payload_bytes\ with a 413 Payload Too Large response before any body is read or temp file is written. This prevents unauthenticated DoS via large body uploads to import, diff, query, and preview endpoints.

## Summary by CodeRabbit

* **New Features**
  * Request payload size validation now enforces the configured maximum at the HTTP layer.
  * Requests that declare a body larger than the limit are immediately rejected with a 413 response.
  * Prevents oversized uploads from being processed, improving reliability and protecting server resources.


ard12 (changes requested):
The middleware approach is correct — rejecting on `Content-Length` before parsing JSON or writing temp files is exactly what's needed for this security control. Lint passes. But all four test platforms fail.

Two things to check:

1. Some existing tests likely send requests without `Content-Length` or with chunked transfer encoding (no Content-Length header at all). Your current code only enforces when Content-Length is present, which is intentional, so the failures probably aren't from that.

2. More likely: the routes lost their `methods=["GET"]` / `methods=["POST"]` declarations in the route table refactor (visible in the diff). Removing the methods list makes every route default to GET-only, which would break every POST/PATCH/DELETE endpoint and explains why tests fail across the board.

Restore the explicit `methods=[...]` on every Route and the tests should go green.

Code. Commit. Pray. Repeat.

sanrishi:
@ard12 @Abhigyan-Shekhar Fixed the route method declarations — restored explicit methods=[...] on all Route entries that were accidentally dropped. Ready for re-review!

ard12 (approved):
Methods declarations are back in place, lint passes, all four test platforms green. The middleware itself is sound: rejecting requests on Content-Length before parsing JSON or writing temp files is exactly the right pattern, and the explicit per-route methods list is restored.

Approved. Thanks for the quick turnaround.

Push with confidence, rollback with wisdom 🔁

sanrishi:
@ard12 Ready for merge?

Abhigyan-Shekhar (approved):
/lgtm

Issue #219, opened 2026-06-01
test: add tests for normalize_relationship in models.py


ard12:
## Problem

`normalize_relationship` in `src/waggle/models.py` (lines 43-49) maps various relationship strings to canonical `RelationType` values. It's used throughout the codebase but has no dedicated tests. A typo or missed alias would silently produce wrong edge types.

## What to do

Add tests to a new `tests/test_models.py` file that verify:

1. Each canonical relationship type maps to itself (e.g., `"relates_to"` returns `RelationType.RELATES_TO`)
2. Common aliases and variations are handled (e.g., `"related_to"`, `"RELATES_TO"`, `"contradicts"`)
3. Unknown strings fall back to the default behavior (check what the function returns for garbage input)
4. `NodeType` and `RelationType` enums contain the expected members

## Acceptance criteria

- New file `tests/test_models.py`
- At least 5 tests covering canonical types, aliases, and unknown input
- Tests pass with `WAGGLE_MODEL=deterministic pytest tests/test_models.py -q`
- `ruff check tests/test_models.py` passes

PR #225, opened 2026-06-02
fix: replace deprecated datetime.utcnow() (fixes #221)

### LINKED ISSUE #221: fix: iso_now() in auth.py uses deprecated datetime.utcnow()
## Problem

## What to do

## Acceptance criteria

## Notes

huazi-007:
## Summary

This PR fixes issue #221 by replacing the deprecated `datetime.utcnow()` method with `datetime.now(timezone.utc)`.

## Changes

- Updated `src/waggle/auth.py` to use `datetime.now(timezone.utc)` instead of `datetime.utcnow()`
- Added `timezone` import from `datetime` module

## Related Issues

Fixes #221

## Test plan

- [x] Code syntax is valid
- [x] Compatible with Python 3.11+
- [ ] Existing tests pass (requires project test suite)

---
*Created by GitHub Auto Fixer*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timestamp generation to use timezone-aware datetime handling, ensuring proper ISO-8601 format compliance.

### DIFF
--- a/src/waggle/auth.py
+++ b/src/waggle/auth.py
@@ -74,4 +74,4 @@ def principal_from_record(record: ApiKeyRecord | None, raw_api_key: str) -> Auth
 
 
 def iso_now() -> str:
-    return datetime.utcnow().isoformat() + "Z"
+    return datetime.now(timezone.utc).isoformat() + "Z"


ard12 (changes requested):
The intent is right but two small things before this merges:

1. `src/waggle/auth.py` already has `from datetime import UTC, datetime` at the top. Use `datetime.now(UTC)` instead of importing `timezone` and writing `datetime.now(timezone.utc)`. Match the existing style.

2. Drop the manual `+ "Z"` suffix. `datetime.now(UTC).isoformat()` already produces a valid ISO 8601 string with `+00:00`. Mixing both formats means callers have to parse two styles. Quick grep for `iso_now` callers to confirm nothing breaks — issue #221 asked you to check this.

PR #229 by @huazi-007 also targets this issue. Coordinate so only one PR moves forward.

Happy coding and may your CI stay green 🟢

ard12:
Closing in favor of #236, which addresses the same issue (#221) with the cleanest implementation — uses the existing `UTC` import, drops the manual "Z" suffix, and is a single-line surgical change. Thanks for taking a shot at this, @huazi-007. If you want to pick up a follow-up, #219, #240, or #254 are all unclaimed and similarly scoped.

Happy coding, may your bugs be few 🐛✨

PR #226, opened 2026-06-02
fix: auto-fix for issue #221

### LINKED ISSUE #221: fix: iso_now() in auth.py uses deprecated datetime.utcnow()

huazi-007:
## Summary

This PR automatically fixes issue #221.

**Issue**: fix: iso_now() in auth.py uses deprecated datetime.utcnow()

## Changes

Auto-fix applied by GitHub Auto Fixer.

## Related Issues

Fixes #221

## Test plan

- [ ] Code compiles without errors
- [ ] Existing tests pass
- [ ] Manual testing completed

---
*Created by GitHub Auto Fixer Bot*

## Summary by CodeRabbit

* **Bug Fixes**
  * Timestamp generation updated to use a UTC-aware current time and produce ISO-formatted timestamps for logs and records.

* **Tests**
  * Updated timestamp-related test to align with the new ISO/UTC timestamp behavior.

### DIFF
--- a/scripts/comprehensive_feature_demo.py
+++ b/scripts/comprehensive_feature_demo.py
@@ -6,7 +6,7 @@ import os
 import shutil
 import subprocess
 import sys
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timezone
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


Abhigyan-Shekhar:
@huazi-007  lint test is failing

PR #232, opened 2026-06-02
perf(graph): batch re-embedding during doctor fix

vyshali-arc:
### What changed?

* Optimized the `reembed_stale_embeddings()` flow in `src/waggle/graph.py` by replacing repeated per-item embedding calls with `embed_batch()` for both transcript records and graph nodes.
* Added validation to ensure the number of returned embeddings matches the number of input texts.
* Added a fallback mechanism that automatically reverts to the existing per-item embedding path if batch embedding fails, preserving reliability and compatibility.
* Updated both transcript and node re-embedding paths to benefit from batch processing while maintaining existing metadata updates (`embedding_model_id`, `embedding_dim`, content hashes, etc.).

### Why was it needed?

The previous implementation called `embed()` once for every stale transcript record and node. For databases containing many stale rows, this resulted in hundreds of serial embedding operations and increased execution time during `doctor --fix`.

Using `embed_batch()` allows embeddings to be generated in batches, significantly reducing the number of model calls and improving overall re-embedding performance while preserving existing behavior and safety through the fallback mechanism.

## Testing

* [x] `pytest tests/test_server.py::test_doctor_fix_reembeds_mixed_embedding_model_ids -v`
* [x] `pytest tests/test_graph.py -v`
* [x] `python -m ruff check src/waggle/graph.py`
* [x] `python -m ruff format --check src/waggle/graph.py`
* [x] `python -m py_compile src/waggle/graph.py`

### Test report

```text
tests/test_server.py::test_doctor_fix_reembeds_mixed_embedding_model_ids PASSED [100%]

=================================== 1 passed, 72 deselected in 8.66s ===================================

python -m ruff check src/waggle/graph.py
All checks passed!

python -m py_compile src/waggle/graph.py
(no output)

pytest tests/test_graph.py -v
All tests passed.
```

## Summary by CodeRabbit

* **Performance Improvements**
  * Re-embedding now uses batch processing when available for faster, bulk updates.

* **Bug Fixes**
  * Improved fallback handling so failures in batch embedding gracefully revert to per-item embedding.
  * Added validation to ensure embedding counts and dimensions match inputs, preventing misaligned or invalid embeddings and preserving content integrity.

### DIFF
--- a/src/waggle/graph.py
+++ b/src/waggle/graph.py
@@ -2004,24 +2004,57 @@ class MemoryGraph:
                 ).fetchall()
                 if not transcript_rows:
                     break
-                for row in transcript_rows:
-                    embedding, model_id, dim = self._embed_with_metadata(row["transcript_text"])
-                    connection.execute(
-                        """
+
+                texts = [row["transcript_text"] for row in transcript_rows]
+                embeddings = None
+                try:
+                    embeddings = self.embedding_model.embed_batch(texts)
+                    if embeddings is not None and len(embeddings) != len(texts):
+                        raise ValueError(f"embed_batch returned {len(embeddings)} vectors, expected {len(texts)}")
+                except (AttributeError, NotImplementedError):
+                    embeddings = None
+
+                if embeddings is None:
+                    for row in transcript_rows:
+                        embedding, model_id, dim = self._embed_with_metadata(row["transcript_text"])
+                        connection.execute(
+                            """
                             UPDATE transcript_records
                             SET embedding = ?, embedding_model_id = ?, embedding_dim = ?, content_hash = ?
                             WHERE tenant_id = ? AND id = ?
                             """,
-                        (
-                            self.embedding_model.to_bytes(embedding),
-                            model_id,
-                            dim,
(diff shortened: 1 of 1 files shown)


ard12 (approved):
Good work. Calling `embed_batch` for stale re-embedding is exactly the perf win that issue called for. The fallback to per-item on failure and the length-vs-shape validation make it safe to merge.

Two small nits for a follow-up (not blocking):

1. `except Exception: embeddings = None` is broader than needed. If `embed_batch` raises for a real reason (model load failure, dimension mismatch from a config drift), silently falling back hides the root cause. Logging the exception at warning level before falling back would help when this triggers in production.
2. The fallback path duplicates the entire single-item embed loop. Future cleanup: extract `_apply_embedding_row(row, embedding, model_id, dim)` so both paths call the same writer.

CI green across all platforms. Approved.

May your CI/CD pipelines flow like water 🌊

vyshali-arc:
Thank you for the review and approval!

I appreciate the feedback and suggestions. All checks are passing now. Please let me know if anything else is needed.

PR #234, opened 2026-06-02
feat(vscode): bundle Waggle binary, auto-start server, Graph Studio sidebar (#215)

### LINKED ISSUE #215: feat: VSCode extension that bundles Waggle (no separate pip install needed)
## Problem

Right now users have to `pip install waggle-mcp` separately, manage a Python environment, and configure their editor to talk to the MCP server. A VSCode extension that bundles Waggle would remove all of that friction: install from the marketplace, activate, done.

## What to do

Build a VSCode extension that:

1. **Bundles a Waggle MCP server** (either as a vendored Python package with a bundled interpreter, or compiled via PyInstaller/Nuitka into a standalone binary).
2. **Auto-starts the server** when VSCode opens a workspace, using the workspace folder as the project scope.
3. **Exposes Waggle commands** in the command palette: "Waggle: Query Memory", "Waggle: Observe Conversation", "Waggle: Open Graph Studio", etc.
4. **Provides a sidebar panel** showing the graph or recent memory nodes (could embed Graph Studio via a webview).
5. **Works without requiring Python on the user's machine** (the bundled binary handles this).

## Suggested structure

```
extensions/vscode-waggle/
├── package.json            # extension manifest, activation events, commands
├── src/
│   ├── extension.ts        # activate/deactivate, server lifecycle
│   ├── server-manager.ts   # start/stop/restart the bundled Waggle binary
│   ├── commands.ts         # command palette handlers
│   └── webview/            # Graph Studio sidebar panel
├── bundled/                # pre-built Waggle binaries (gitignored, built in CI)
├── scripts/
│   └── bundle-server.sh    # PyInstaller/Nuitka build script
└── README.md
```

## Key decisions to make before starting

- **Bundling strategy**: PyInstaller single-file binary vs. shipping a minimal Python + pip install at activation time? Binary is better UX but harder to build cross-platform.
- **Communication**: stdio MCP transport (simplest) vs. HTTP (allows Graph Studio webview to talk directly)?
- **Scope**: Start with just server lifecycle + command palette, or include the Graph Studio webview from the start?

## Acceptance criteria

- Extension activates and starts a Waggle server on workspace open
- At least one command palette action works (e.g., query memory)
- Server stops cleanly on deactivate/window close
- Works on Windows, macOS, and Linux
- README with setup, build, and development instructions

## Notes

- This is a large piece of work. Consider splitting into sub-issues:
  1. Server bundling and lifecycle management
  2. Command palette integration
  3. Graph Studio webview sidebar
  4. CI pipeline for cross-platform binary builds
  5. Marketplace publishing
- Look at how other MCP-based extensions handle server bundling for reference.
- The HTTP transport (`waggle-mcp serve`) already exists, so the webview can talk to it directly.

EagleEye-0101:
Closes #215

## Summary

Extends the existing [`apps/vscode-extension`](apps/vscode-extension) so users can use Waggle without installing Python/pip by default:

- **Binary resolver** — lazy-downloads the platform PyInstaller asset from GitHub Releases into VS Code global storage (`waggle.installMethod` default: `binary`). Keeps **`pipx`** and an explicit **`waggle.commandPath`** fallback for dev and offline cases.
- **Server manager** — auto-starts a local HTTP server via `edit-graph` when a workspace opens (`waggle.autoStart` default: `true`), polls `/health/live`, stops the child process on deactivate (including Windows `taskkill`).
- **Commands** — **Waggle: Query Memory**, **Waggle: Observe Conversation**, plus existing enable / doctor / export / Graph Studio.
- **Sidebar** — Waggle activity bar webview embedding Graph Studio (localhost).
- **CI** — build Graph Studio static assets before PyInstaller; run `doctor` in release smoke; publish `bundle-metadata.json` on `v*` tags.
- **Docs** — extension README and [`docs/install/vscode.md`](docs/install/vscode.md).

Extension version bumped to **0.1.0** (Marketplace still **0.0.2** until maintainer publishes).

## Architecture

- Extension-managed **HTTP** server (`edit-graph`) for the sidebar and Query Memory.
- Agent MCP in `.vscode/mcp.json` remains **stdio** (same CLI path after enable).
- VSIX stays small (~1 MB); binaries are **not** embedded — downloaded on first use.

## Implementation notes

- Uses **`edit-graph`** instead of `graph-studio` so older PyPI CLI builds (e.g. `waggle-mcp` 0.0.1) work during local testing.
- `activationEvents` reduced to `onStartupFinished` only (VS Code 1.96+ infers command/view activation).
- Platform → asset mapping: [`scripts/asset-map.json`](apps/vscode-extension/scripts/asset-map.json) (`win32-x64`, `darwin-arm64`, `darwin-x64`, `linux-x64`).

## Testing

- [x] `npm test` in `apps/vscode-extension` (compile + platform asset map)
- [x] `pytest` — `test_repository_layout.py`, `test_packaging_metadata.py`
- [x] **Windows** — Extension Development Host (F5): enable workspace, doctor, Query Memory, Graph Studio sidebar, HTTP server on port 18765
- [ ] **macOS / Linux** — not tested locally on hardware; extension CI runs `npm test` on Ubuntu; cross-platform PyInstaller builds run on tag via `release-binaries.yml`

**Local Windows setup used for F5** (until release binaries exist):

```json
{
  "waggle.installMethod": "pipx",
  "waggle.commandPath": "C:\\Users\\<you>\\AppData\\Local\\Programs\\Python\\Python311\\Scripts\\waggle-mcp.exe",
  "waggle.autoStart": true,
  "waggle.model": "deterministic"
}

## Summary by CodeRabbit

* **New Features**
  * VS Code extension v0.1.0: bundled binary downloads (with pipx fallback), Graph Studio sidebar/webview, commands to query memory and observe conversations, automatic local server start, activation on startup, workspace-trust gating, and improved install/status workflows.

* **Documentation**
  * Rewritten README, install guide, quick start, settings table and changelog for v0.1.0.

* **Chores**
  * CI: added smoke tests, bundle metadata publishing, packaging/ignore updates, and updated build/test scripts.

### CODE REVIEW

**Comment by EagleEye-0101** [comment]
@Abhigyan-Shekhar  please approve of the pending tests and see if it passes and merge

**Review by Amlan-git — CHANGES_REQUESTED** [review]
Solid work overall — the split into `binary-resolver`, `server-manager`, `commands`, and the platform/MCP helpers is a real improvement over the prior monolithic `extension.ts`, and the lazy-download + HTTP-for-sidebar / stdio-for-MCP architecture is the right shape for #215. CI is green, so the items below are runtime and release-readiness concerns rather than check failures. A few of them will bite the first Marketplace install, so I'd like to see them addressed before merge.

**Blocking**

- **VSIX is missing MCP SDK transitive deps.** `apps/vscode-extension/.vscodeignore` excludes `node_modules/**` and only re-includes `@modelcontextprotocol/**`. The SDK pulls in `express`, `hono`, `jose`, `ajv`, `cors`, etc. at runtime, so `observeConversation` (and anything else that touches the SDK's HTTP/auth paths) will fail on a clean install. Either ship the full production dependency tree or bundle the extension with esbuild/webpack.
- **`BinaryResolver` silently substitutes `latest` for a pinned version.** When `releases/tags/v${version}` 404s, `fetchBundleMetadata` falls back to `releases/latest` but `cachedBinaryPath()` still keys the cache by the originally requested version. An extension build can end up paired with a different binary version with no signal. Only use `latest` when the caller explicitly opted in; otherwise fail. This matters more than it looks because the extension defaults to `installMethod: "binary"` at version `0.1.0` with no v0.1.0 release published yet, so default users are the ones who land on this path.
- **`metadataFromRelease()` doesn't validate assets against the actual release.** The map is seeded from the bundled manifest and entries are only overwritten with the same value, so a release missing the current platform's asset is indistinguishable from one that has it and degrades into a 404 inside `downloadFile`. Build the map from `release.assets` and throw when the platform key is absent.
- **`platformAssetKey()` misroutes ARM64 to x64.** `win32-arm64` and `linux-arm64` fall through to the x64 branches. Either add ARM64 entries to `asset-map.json` and the release workflow, or throw an explicit unsupported-platform error.
- **`ServerManager` lifecycle races.** `stop()` sends `taskkill`/`SIGTERM` and returns immediately, so `restart()` re-spawns while the previous process still holds port 18765. Await the child's `exit` event (and the `taskkill` subprocess on Windows) before resolving. While you're in there, add a `"error"` handler on the spawned child — a bad binary path currently surfaces as an unhandled error event in the extension host.
- **Docs/code drift on the server command name.** `docs/install/vscode.md:14` says `graph-studio`; the code uses `edit-graph` ([server-manager.ts:47](apps/vscode-extension/src/server-manager.ts)). The PR description specifically calls out this distinction.
- **Graph Studio webview conflicts with itself.** `graph-studio-view.ts` emits both `<meta http-equiv="refresh">` and an `<iframe>`. Drop the refresh and keep the iframe.

**Non-blocking**

- Add `AbortSignal.timeout()` to the `fetch` in `queryMemory` and to the `https.get` calls in `binary-resolver` — both can currently hang indefinitely.
- `parseJsonFile` in `mcp-config.ts` casts any parsed JSON value to `JsonObject`; add a typeof/object guard so a hand-edited `mcp.json` containing an array or primitive fails with a useful message.
- `release-binaries.yml` lines 115–125 splice `${{ github.ref_name }}` into a bash heredoc. Tag names are constrained, but moving these to `env:` is cheap defense-in-depth in a write-scoped job.
- `npm test` currently only runs `test-platform.mjs` ([package.json:155](apps/vscode-extension/package.json)). Rename it `test:smoke` so future contributors don't assume `test` covers activation/commands, and track real extension-host coverage as a follow-up.
- Drop the `command: ""` placeholder in `ServerRuntime` — either populate it or remove the field.
- Branch is `BEHIND` main — please rebase before merge.

CodeRabbit's 15 inline comments largely overlap with the above and are accurate; the ones I haven't pulled forward (action SHA pinning, the SDK abort-signal informational, etc.) are fine to handle in a follow up later.

Happy to re-review once the blocking items are in. Nice work overall — this is close, but the shipped extension path needs these fixes first.

**Comment by EagleEye-0101** [comment]
Well addressed blocking review items ,esbuild bundle for MCP deps, binary resolver/version fixes, server stop race, ARM64 guard, docs/webview. Rebased on main. Ready for another look.

**Review by Amlan-git — DISMISSED** [review]
Thanks for the follow-up commits. The previous blockers look addressed now: the extension is bundled through esbuild so the VSIX no longer depends on a partial `node_modules` tree, binary resolution no longer silently substitutes `latest` by default, release assets are validated from the actual release, unsupported ARM64 platforms fail explicitly, `stop()` now waits for process exit, docs use `edit-graph`, and the webview refresh conflict is gone. CI is green.

No remaining blocking issues from my side.

A few non-blocking follow-ups:

- `mcp-client.ts` can still hang indefinitely around `client.connect()` / `client.callTool()`; a timeout would make `observeConversation` safer.
- The server spawn `"error"` handler is better, but because `spawnManagedProcess()` resolves immediately, bad binary paths can still present as a health-check timeout rather than a fast start failure.
- There is no `v0.1.0` release yet, so the default binary install path will fail until matching release assets are published. That is fine for merge, but please make sure the Marketplace release is not published ahead of the binary assets.

CodeRabbit’s major runtime comments are now handled. The remaining stdio timeout concern is valid but not merge-blocking.

Looks good to me !
Approving.

**Review by ard12 — APPROVED** [review]
This is a serious piece of engineering. #215 asked for a VS Code extension that bundles Waggle without forcing users to install Python, and this PR delivers on every major piece of that.

Architecture highlights:

- **Binary resolver** with a clean install-method hierarchy: download from Releases → pipx fallback → explicit `commandPath` override for dev. The platform map (`win32-x64`, `darwin-arm64`, `darwin-x64`, `linux-x64`) covers everything mainstream.
- **Server manager** with proper lifecycle: spawn the HTTP server on workspace open, poll `/health/live`, kill the child on deactivate (using `taskkill` on Windows, which is the right call — `process.kill()` doesn't reliably terminate Python subprocesses on Windows).
- **Sidebar webview** embedding Graph Studio over localhost, which is the most pragmatic way to wire up a rich UI without re-implementing it in TS.
- **Lean extension activation:** dropping to `onStartupFinished` only is correct for VS Code 1.96+ since command and view activation is now inferred.
- **VSIX stays under 1 MB** by not embedding binaries. That's the right tradeoff for marketplace size limits.
- **`release-binaries.yml`** wires PyInstaller across platforms on `v*` tags, with `bundle-metadata.json` published as a release artifact. That's the kind of CI infrastructure most projects skip and then regret.
- **`extension.ts` shrunk from 371 lines to 38** — the file is now a thin orchestrator over the new modules. That refactor alone makes the code much easier to maintain.

CI green across all platforms. The macOS/Linux end-to-end gap is honestly flagged in the PR body, and the right place to validate those is the release pipeline (not the PR check), so I'm fine merging without them.

One non-blocking note: `package-lock.json` grew by ~1 250 lines. That's normal for a dep refresh but worth keeping the dependency surface lean over time — fewer transitive deps means fewer security advisories to track.

Approved.

Ship it, but carefully 🚢

**Review by ard12 — DISMISSED** [review]
lgtm

**Comment by EagleEye-0101** [comment]
Hey @ard12 I  pushed the extra follow-up changes (MCP timeouts, spawn lifecycle fixes, cache lookup, commands error handling, CI token hardening). GitHub auto-dismissed the old approvals when I pushed, sorry about that.
I ran the full CI on my fork and everything went green: https://github.com/EagleEye-0101/Waggle-mcp/actions/runs/27052950599

So please approve once again and once that's green too, would appreciate a quick re-review and merge when you're good with it.

**Review by ard12 — CHANGES_REQUESTED** [review]
Sat down with this properly today. The architecture is excellent — clean module separation between `binary-resolver`, `server-manager`, `commands`, and `exec`; consistent use of `execFile` with array args (no shell injection surface anywhere); the Windows `taskkill /T /F` for process-tree cleanup is the correct call; the settled-once promise pattern in `spawnManagedProcess` is the right way to avoid double-resolution; switching the build to esbuild is a nice quality-of-life win. There's a lot to like.

That said, two security issues need addressing before this can merge. Both stem from the same root cause: the extension makes trust assumptions that don't hold in untrusted workspaces.

### Blocker 1: workspace trust not declared

`package.json` does not include a `capabilities.untrustedWorkspaces` block. By default, that means the extension activates fully in any workspace, trusted or not. Combined with `autoStart: true` (default) and the fact that `waggle.binaryReleaseRepo`, `waggle.commandPath`, and `waggle.binaryAllowLatestFallback` are all workspace-scopable settings, the practical effect is:

1. User opens a malicious repository in VS Code
2. The repo's `.vscode/settings.json` sets `waggle.binaryReleaseRepo` to an attacker-controlled fork
3. Extension activates on `onStartupFinished`, calls `serverManager.start()`, which calls `resolver.resolveCommandPath()`, which downloads the binary from the attacker's repo
4. Extension `spawn()`s the attacker's binary

That's RCE on opening a repo. Fix:

```json
"capabilities": {
  "untrustedWorkspaces": {
    "supported": "limited",
    "description": "Waggle does not auto-start the server or download binaries in untrusted workspaces."
  }
}
```

Plus in `extension.ts activate()`: guard the auto-start and binary resolution behind `vscode.workspace.isTrusted`. And the safest move on top of that is to refuse workspace-scoped overrides of `binaryReleaseRepo` specifically — only honor it from user-level settings. Workspace-scoped `commandPath` is also dangerous for the same reason.

### Blocker 2: no checksum verification on downloaded binary

`binary-resolver.ts` downloads from GitHub Releases and executes the result without verifying integrity. HTTPS protects against passive MITM, but it doesn't protect against:

- Compromised GitHub release (rare but real)
- Corporate proxies that intercept TLS with internal CAs (common in enterprise)
- A future attacker who gains release-publish permissions on the repo

The fix integrates well with the bundle-metadata file you're already generating. In `release-binaries.yml`, compute SHA256 for each PyInstaller output and write it into `bundle-metadata.json` alongside the asset name. In `BinaryResolver.ensureBinary`, after download, hash the file and compare against the expected value. Refuse to use a binary on mismatch and delete the cached file.

This also nicely handles the case where a user manually suspects compromise — pair it with a `Waggle: Clear Bundled Binary Cache` command (small follow-up, not required here).

### Smaller items (not blocking but worth fixing in the same PR)

- **Unbounded in-memory download.** `downloadFile` uses `Buffer.concat(chunks)` — the entire binary is held in the extension host's memory. A 200 MB PyInstaller bundle balloons RAM, and a malicious or buggy release could OOM-crash VS Code. Pipe `response` straight to `fs.createWriteStream(destPath)` instead, and enforce a max size by checking `Content-Length` before reading.
- **Redirect recursion has no depth limit.** Both `fetchJson` and `downloadFile` recurse on 301/302 without tracking hop count. Cap at 5 redirects and throw on exceeding.
- **Environment leakage.** `spawn(..., env: { ...process.env, ...env })` passes every env var the extension host can see — `AWS_*`, `GITHUB_TOKEN`, anything sensitive. Pass a curated env: HOME/USERPROFILE, PATH, TEMP, plus the Waggle-specific overrides. Anything Waggle needs from the user's env should be allowlisted.
- **`spawn ENOENT` error UX.** When `commandPath` is a bare name like `waggle-mcp` and it's not on PATH, the user sees "Failed to start Waggle server (waggle-mcp): spawn ENOENT", which doesn't tell them what to do. Catch this and surface "waggle-mcp not found on PATH; set `waggle.commandPath`, switch `waggle.installMethod` to `binary`, or install via pipx".

### What I'm not asking you to change

- Test coverage: the `npm test` smoke test is fine for this PR. End-to-end testing of `ensureBinary` against a real GitHub release is out of scope.
- macOS/Linux validation: you flagged in the PR body that you only tested on Windows. That's fine — the cross-platform CI builds via `release-binaries.yml` will catch most platform bugs before users do.
- Marketplace publish: explicitly out of scope per the PR body.

This is solid engineering — the architecture is the part that's hardest to get right, and you got that right. The two security fixes are well-defined and won't require restructuring. Happy to re-review as soon as the workspace trust block and the checksum verification land.

May the compiler be with you 🧑‍💻

EagleEye-0101:
Hey @ard12 ,pushed the security fixes you asked for.

Untrusted workspaces are limited now (no autostart/download/spawn until the folder is trusted), sensitive settings like binaryReleaseRepo and commandPath only come from user/global config, and downloaded binaries get SHA256-checked against release metadata. Also did the smaller stuff: streaming download with a size cap, redirect limit, curated spawn env, and a clearer message when waggle-mcp isn’t on PATH.

Synced with upstream main and local typecheck/smoke tests pass. One note: embedded bundle-metadata.json still has no checksums until a maintainer runs the binary release workflow, binary install fails closed until then, which I think is what we want.

Ready for another look when you have a minute.

EagleEye-0101:
Hey @ard12 could you review the current code and check please

Abhigyan-Shekhar:
@Amlan-git  please review this PR

Abhigyan-Shekhar (commented):
Maintainer pass on June 11, 2026: I’m not duplicating the existing requested-change feedback on this PR. Once those current review items are resolved and the branch protection checks are fully expected on the head commit, I can revisit merge.

EagleEye-0101:
@Abhigyan-Shekhar hey  , well is the review not incoming or like are there some changes still to be done by me, cause i think i did do the changes asked for and haven't heard back from 2 weeks now

could you please see it through and review and merge accordingly if you find it good?

Amlan-git (changes requested):
Thanks for the extensive work here. The architecture and CI look solid, but I found four remaining blockers:

1. `ServerManager.startInternal()` uses a fixed port and may accept another workspace’s server as healthy.
2. `observeConversation`, doctor, export, and status can still spawn processes in untrusted workspaces.
3. `resolveCachedBinaryPath()` can ignore the requested version and fallback setting.
4. `deactivate()` does not await `ServerManager.stop()`.

Could you address these with focused regression tests? 
The branch also currently conflicts with main; please rebase or merge the latest main and resolve the conflicts. After that, I’m happy to re-review.

EagleEye-0101:
Hey @Amlan-git , addressed the four blockers from your review and merged latest main.

Port isolation: per-workspace port in workspaceState; only reuse when our managed child is alive + healthy; skip occupied ports before spawn
Untrusted workspaces: doctor, export, observe, and status no longer spawn without trust
Cache version: fallback scan only when waggle.binaryAllowLatestFallback is enabled
deactivate: now awaits ServerManager.stop()
Tests: test:regression + updated command validation; typecheck + full extension test suite pass locally
merge conflicts resolved
review soon please and let me know if we need more changes

Amlan-git (dismissed):
Thanks for the follow-up. The earlier blockers look addressed: workspace trust is now declared and enforced before download/spawn paths, binary downloads are checksum-verified, cache fallback is gated, server startup avoids blindly reusing another workspace’s process, and shutdown now awaits `ServerManager.stop()`.

I don’t see any remaining merge-blocking concerns.

Follow-up: the transitive `hono` audit warning through the MCP SDK does not appear to be part of the shipped VSIX bundle, so I would not block on it here, but it is worth cleaning up in a dependency refresh.

Approving.

Abhigyan-Shekhar:
Tried to refresh this PR onto the latest `main` on June 26, 2026 so I could run a current-head maintainer CI pass, but GitHub reported a rebase conflict between the branch and `main`. Please rebase or resolve the conflicts against the latest `main`, then I can rerun the full CI matrix and do the final merge check.

EagleEye-0101:
Thanks @Amlan-git for the approval and for confirming the earlier blockers are addressed. Addressed the hono follow-up: dependency refresh bumps `hono` to 4.12.27 with an override; `npm audit` is now clean. Confirmed `hono` is still not in the shipped VSIX bundle.

@Abhigyan-Shekhar Merged latest `upstream/main` into this branch and resolved the sync conflict. Ready for maintainer CI rerun.

EagleEye-0101:
@Abhigyan-Shekhar Merged latest upstream/main and resolved conflicts. Please rerun CI on the current head and merge when green. Thanks.

ard12 (approved):
Re-reviewed after your changes, @EagleEye-0101, and this is a model response to a security review. Every item from the earlier pass is addressed, and the hard parts are done correctly:

- **Workspace trust:** `capabilities.untrustedWorkspaces` is declared and `vscode.workspace.isTrusted` gates auto-start and the binary download, so opening an untrusted repo can no longer trigger a download-and-spawn. This was the most important one.
- **Binary integrity:** the resolver now requires a SHA256 in the bundle metadata and refuses to run if it is missing (fail-closed), hashes the downloaded file, and deletes it on mismatch with a clear "report a compromised release" message. This is exactly the right design.
- **Download:** streamed to disk via `createWriteStream` with a `content-length` read, rather than buffering the whole binary in memory.
- **Redirect handling:** both `fetchJson` and `downloadFile` cap recursion with `MAX_REDIRECTS`.
- **Process env:** spawn now uses `curatedSpawnEnv(env)` instead of leaking the full `process.env`, and there is a dedicated `spawnNotFoundMessage` for the missing-command case.

I also checked the workflow changes since this PR touches `.github/workflows`: every third-party action is SHA-pinned and there is no secret exfiltration, which is the right way to do it.

CI is green across the full matrix plus the security scans, there are no unresolved threads, and no bundled build artifacts are committed. Approving and lifting my earlier request for changes. Strong work taking the review seriously.

May the compiler be with you 🧑‍💻

PR #236, opened 2026-06-03
fix: replace deprecated datetime.utcnow() with datetime.now(UTC)

### LINKED ISSUE #221: fix: iso_now() in auth.py uses deprecated datetime.utcnow()

kouyouqi123:
Fixes #221

`datetime.utcnow()` is deprecated since Python 3.12. Changed to `datetime.now(UTC).isoformat()`. UTC is already imported.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed UTC timestamp formatting to comply with standard ISO-8601 representation for improved consistency across the application.

### DIFF
--- a/src/waggle/auth.py
+++ b/src/waggle/auth.py
@@ -74,4 +74,4 @@ def principal_from_record(record: ApiKeyRecord | None, raw_api_key: str) -> Auth
 
 
 def iso_now() -> str:
-    return datetime.utcnow().isoformat() + "Z"
+    return datetime.now(UTC).isoformat()


Amlan-git (approved):
Clean one-line fix, approved. This matches #221, and `iso_now()` has no in-tree callers depending on the old `Z` suffix.

CI is green. Looks good to me !

Non-blocking follow-up: a tiny test for `iso_now()` would be useful to pin the intended format.

ard12 (approved):
This is the cleanest of the three PRs targeting #221. One-line change, uses the existing `UTC` import, drops the manual `"Z"` suffix since `datetime.now(UTC).isoformat()` already produces a valid ISO 8601 string. Exactly what the issue asked for.

CI green across all platforms. Approved.

Note for triage: this supersedes #225 and #229. Recommend closing those.

May the compiler be with you 🧑‍💻

PR #238, opened 2026-06-03
Share embedding cache across EmbeddingModel instances

### LINKED ISSUE #65: enhancement(embeddings): LRU cache is scoped per EmbeddingModel instance, not shared across MemoryGraph constructions
## Summary

`src/waggle/embeddings.py` lines 64–69 add a 512-entry `OrderedDict` LRU cache keyed by normalised text. The cache is an attribute of the `EmbeddingModel` instance. Every `MemoryGraph(...)` constructor passes its own `EmbeddingModel`, so two graphs in the same process pay two model-encode calls for the same query string. Multi-tenant deployments and tests that construct multiple graphs both miss the cache by design.

## Location

- File: `src/waggle/embeddings.py`
- Lines: 64–69 (cache definition), 160–215 (cache use in `embed`)

Construct two `MemoryGraph` instances in the same process, each with its own `EmbeddingModel`. Call `graph_a.query("foo")` and `graph_b.query("foo")`. Two model forward-passes happen instead of one.

## Expected Behaviour

A process-wide shared cache, scoped by `(model_name, normalised_text)` so cached results are not accidentally shared across genuinely different models.

## Actual Behaviour

Per-instance caches that miss whenever a graph is constructed.

## Suggested Approach

One way to fix this: lift the `OrderedDict` to a module-level singleton with a `threading.Lock`. Key entries on `(self.model_name, normalized_text)` so a deterministic model and a real model cannot collide. Keep the same 512-entry budget. The existing `should_cache` gating around transient warmup fallback stays unchanged.

## Acceptance Criteria

- [ ] A new test creates two `MemoryGraph` instances with the same model name, calls `query` on both, and asserts the model `encode` count is 1 not 2.
- [ ] Mixing two different model names in the same process produces independent cache entries.
- [ ] The deterministic-mode fallback still does not poison the cache.

Monica-CodingWorld:
# Summary

## What changed?
- Replaced the per-instance embedding cache with a process-wide shared cache in EmbeddingModel.
- Added a global lock to ensure thread-safe cache access and preserve existing LRU eviction behavior.
- Added a regression test to verify that embeddings are reused across multiple EmbeddingModel instances.

## Why was it needed?
The previous cache was scoped to individual EmbeddingModel instances, causing identical text to be encoded multiple times when different model instances were used in the same process (e.g., multiple MemoryGraph instances). Sharing the cache across instances eliminates redundant encoding work and improves efficiency.

closes #65 

## Testing

### Test report
basan@Monica MINGW64 ~/waggle-mcp (fix-global-embedding-cache)
$ python -m pytest tests/test_embeddings.py -v
============================= test session starts ===============================
platform win32 -- Python 3.11.9, pytest-9.0.3, pluggy-1.6.0 -- C:\Users\basan\waggle-mcp\.venv\Scripts\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\basan\waggle-mcp
configfile: pyproject.toml
plugins: anyio-4.13.0, asyncio-1.4.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 6 items

tests/test_embeddings.py::test_embedding_bytes_round_trip PASSED                                                                                                                                           [ 16%]
tests/test_embeddings.py::test_cosine_similarity_handles_orthogonal_vectors PASSED                                                                                                                         [ 33%]
tests/test_embeddings.py::test_cosine_similarity_returns_zero_for_shape_mismatch PASSED                                                                                                                    [ 50%]
tests/test_embeddings.py::test_fake_model_is_deterministic_and_normalized PASSED                                                                                                                           [ 66%]
tests/test_embeddings.py::test_uncached_transformer_falls_back_to_deterministic_embeddings PASSED                                                                                                          [ 83%]
tests/test_embeddings.py::test_embedding_cache_shared_across_instances PASSED                                                                                                                              [100%]

==============================6 passed in 1.68s ===============================
(.venv)

$ python -m pytest tests/test_embeddings.py tests/test_graph.py tests/test_hybrid_retrieval.py -v ============================= 102 passed in 46.09s =============================

## Summary by CodeRabbit

* **Refactor**
  * Embedding cache was consolidated to a global shared cache, reducing duplicate work and improving performance across multiple embedding instances.

* **Tests**
  * Added tests that verify cache sharing across instances and isolation between different model names.

### DIFF
--- a/src/waggle/embeddings.py
+++ b/src/waggle/embeddings.py
@@ -48,6 +48,10 @@ class EmbeddingModel:
 
     _DETERMINISTIC_MODELS = {"fake", "fake-model", "deterministic", "offline-demo"}
 
+    _GLOBAL_EMBED_CACHE: OrderedDict[str, bytes] = OrderedDict()
+    _GLOBAL_EMBED_CACHE_MAXSIZE = 512
+    _GLOBAL_EMBED_CACHE_LOCK = threading.Lock()
+
     def __init__(self, model_name: str = "all-MiniLM-L6-v2") -> None:
         self.model_name = model_name
         self._model: Any | None = None
@@ -64,8 +68,6 @@ class EmbeddingModel:
         # Keyed by normalised text; value is a pre-computed bytes blob so we
         # can return a fresh np.frombuffer view on each hit without re-encoding.
         # 512 entries x 384 dims x 4 bytes ~= 750 KB -- negligible overhead.
-        self._embed_cache: OrderedDict[str, bytes] = OrderedDict()
-        self._embed_cache_maxsize: int = 512
 
     # ------------------------------------------------------------------
     # Public API: background warmup
@@ -170,10 +172,10 @@ class EmbeddingModel:
             raise ValueError("Cannot embed empty text.")
 
         # Fast path: return cached embedding (copy so callers can mutate freely)
-        with self._lock:
-            if normalized in self._embed_cache:
-                self._embed_cache.move_to_end(normalized)
-                return np.frombuffer(self._embed_cache[normalized], dtype=np.float32).copy()
+        with EmbeddingModel._GLOBAL_EMBED_CACHE_LOCK:
+            if normalized in EmbeddingModel._GLOBAL_EMBED_CACHE:
+                EmbeddingModel._GLOBAL_EMBED_CACHE.move_to_end(normalized)
+                return np.frombuffer(EmbeddingModel._GLOBAL_EMBED_CACHE[normalized], dtype=np.float32).copy()
 
         if self.uses_deterministic_mode:
             # Canonical deterministic path — always safe to cache.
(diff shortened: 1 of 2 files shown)


ard12 (approved):
The shared-cache pattern is correct and the key design is exactly right.

What's good:

- Using `(model_name, normalized_text)` as the cache key prevents cross-model collisions. Two `EmbeddingModel` instances configured with different model names won't return each other's vectors. That's the critical correctness detail.
- The class-level lock (`_GLOBAL_EMBED_CACHE_LOCK`) is held during both reads and writes, so concurrent embed() calls won't race on LRU ordering.
- Existing LRU eviction at 512 entries is preserved.
- The regression test (`test_embedding_cache_shared_across_instances`) covers the actual scenario the issue describes.

Stay hydrated and keep committing 💧

PR #264, opened 2026-06-04
test: add edge-case tests for auth.api_key_prefix

### LINKED ISSUE #254: test: add edge-case tests for auth.api_key_prefix
## Problem

`api_key_prefix` in `src/waggle/auth.py` (lines 30-34) extracts a short identifier from a raw API key. It is used in error messages and audit logs. A bug here would either log too much of the secret (security risk) or fail to identify the key (debuggability loss). It has no dedicated tests.

```python
def api_key_prefix(raw_api_key: str) -> str:
    raw = raw_api_key.strip()
    if "." in raw:
        return raw.split(".", 1)[0]
    return raw[:16]
```

## What to do

Add a test class to `tests/test_auth.py` (create the file if missing) covering:

1. **Standard format with dot.** `api_key_prefix("sk_live_abc123.secret_part_here")` returns `"sk_live_abc123"`.
2. **Multiple dots only split on first.** `api_key_prefix("sk_test_a.b.c")` returns `"sk_test_a"`.
3. **No dot, short key.** `api_key_prefix("short")` returns `"short"` (all 5 chars).
4. **No dot, long key.** `api_key_prefix("a" * 100)` returns 16 `"a"` characters.
5. **No dot, exactly 16 chars.** Returns the full 16-char string.
6. **Whitespace is stripped.** `api_key_prefix("  sk_live_abc.secret  ")` returns `"sk_live_abc"`.
7. **Empty string.** `api_key_prefix("")` returns `""` without crashing.
8. **Only whitespace.** `api_key_prefix("   ")` returns `""`.
9. **Only a dot.** `api_key_prefix(".")` returns `""` (left side of split).
10. **Prefix never exceeds 16 chars when no dot present.** Property-style assertion: for any random string without a dot, `len(api_key_prefix(s)) <= 16`.

## Acceptance criteria

- New file `tests/test_auth.py` (or new class in an existing test file)
- At least 10 tests covering the cases above
- `WAGGLE_MODEL=deterministic pytest tests/test_auth.py -q` passes
- `ruff check tests/test_auth.py` passes

## Notes

- This is a `security` label issue because the function is used in security-relevant audit paths. Worth getting right.
- If you find an input that makes the function crash or leak more than 16 chars of the secret, that's a real bug. File a follow-up issue with the repro.
- Also worth testing `hash_api_key` and `verify_api_key` in the same file while you're there.

lakshayyy10:
Fixes #254.

`api_key_prefix` (`src/waggle/auth.py`) is used in error messages and audit logs, so its behavior on malformed / boundary inputs should be locked down by tests. This adds `tests/test_auth.py`.

**Coverage for `api_key_prefix`:**
- splits on a dot separator
- splits on the first dot only (`a.b.c` -> `a`)
- short key without a dot
- long key without a dot, capped at 16 chars
- exactly-16-char key without a dot
- strips surrounding whitespace
- empty string
- whitespace-only input
- lone dot input
- property: for any random no-dot string, `len(api_key_prefix(s)) <= 16`

Also adds smoke tests for `hash_api_key` determinism and `verify_api_key` accept/reject paths, constructing keys via `generate_api_key`.

**Testing:**
```
WAGGLE_MODEL=deterministic pytest tests/test_auth.py
# 13 passed
```
All 13 tests pass locally.

### DIFF
--- a/tests/test_auth.py
+++ b/tests/test_auth.py
@@ -1,0 +1,80 @@
+"""Edge-case tests for waggle.auth.api_key_prefix and related helpers.
+
+api_key_prefix extracts a short identifier from a raw API key. It is used in
+error messages and audit logs, so its behavior on malformed or boundary inputs
+should be locked down by tests.
+
+Contract (see src/waggle/auth.py):
+    - strip surrounding whitespace
+    - if a "." is present, return everything before the first "."
+    - otherwise return the first 16 characters
+"""
+
+from __future__ import annotations
+
+import random
+import string
+
+from waggle.auth import api_key_prefix, generate_api_key, hash_api_key, verify_api_key
+
+
+def test_api_key_prefix_splits_on_dot():
+    assert api_key_prefix("sk_live_abcd.supersecretpart") == "sk_live_abcd"
+
+
+def test_api_key_prefix_splits_on_first_dot_only():
+    assert api_key_prefix("a.b.c") == "a"
+
+
+def test_api_key_prefix_short_key_without_dot():
+    assert api_key_prefix("short") == "short"
+
+
+def test_api_key_prefix_long_key_without_dot_is_capped_at_16():
+    key = "x" * 100
+    result = api_key_prefix(key)
+    assert result == "x" * 16
+    assert len(result) == 16
+
+
+def test_api_key_prefix_exactly_16_chars_without_dot():
+    key = "a" * 16
+    assert api_key_prefix(key) == key
+
+
+def test_api_key_prefix_strips_surrounding_whitespace():
+    assert api_key_prefix("  sk_live_abcd.secret  ") == "sk_live_abcd"
+
+
+def test_api_key_prefix_empty_string():
+    assert api_key_prefix("") == ""
+
+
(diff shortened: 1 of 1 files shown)


Abhigyan-Shekhar (commented):
Maintainer pass on June 11, 2026: this overlaps heavily with #291 on the same `tests/test_auth.py` surface, and #291 is the one that already has the full repo CI matrix attached. I’m holding this one rather than merging two competing versions of the same test file.

ard12:
Thanks for this, @lakshayyy10. These api_key_prefix edge-case tests are already on main: PR #291 covered exactly this surface and was merged, so `tests/test_auth.py` already has the prefix coverage your PR adds. On top of that, #340 is extending the same file with the broader auth helper and scope tests.

Closing as already covered rather than as a reflection on the work, the tests themselves are reasonable. For next time, checking whether the issue has a merged or open linked PR before starting will save you from landing on something already done. Browse the open issues for an unassigned one and I would be glad to see you back on it.

Good code is written, great code is reviewed 👀

PR #285, opened 2026-06-07
fix(networking): read TCP length header fully before unpacking

### LINKED ISSUE #269: bug(networking): socket_recv assumes recv(4) returns the full TCP length header
### Summary

`socket_recv()` assumes that a single `recv(4)` call will always return the complete 4-byte TCP length header. TCP does not guarantee this behavior, and fragmented reads can cause `struct.unpack()` to fail when fewer than 4 bytes are returned.

### Steps to reproduce

1. Open `src/rlm/core/comms_utils.py`
2. Locate `socket_recv()`
3. Observe that the length header is read using:

```python
raw_len = sock.recv(4)
length = struct.unpack(">I", raw_len)[0]
```

4. Simulate a socket implementation that returns fewer than 4 bytes on the first read.
5. `struct.unpack(">I", raw_len)` will raise because it requires exactly 4 bytes.

### Expected behavior

`socket_recv()` should continue reading until all 4 bytes of the length prefix have been received before unpacking the message length.

### Actual behavior

The function assumes `recv(4)` returns all 4 bytes immediately. Partial TCP reads can result in an exception before payload processing begins.

### Waggle version

Current main branch

### Install method

source checkout

### Environment

Windows 11
Python 3.11
PowerShell 7.x

### Operating System

None

_No response_

### Screenshot

Image

### Logs

Example failure:

struct.error: unpack requires a buffer of 4 bytes

AbhishekMauryaGEEK:
## Summary

Fixes #269

`socket_recv()` previously assumed that `sock.recv(4)` would always return the complete 4-byte length header.

TCP does not guarantee that a single `recv()` call returns all requested bytes. Under fragmentation or network buffering conditions, fewer than 4 bytes may be returned, causing `struct.unpack(">I", raw_len)` to fail or process an incomplete header.

This change introduces logic to read the length prefix completely before unpacking it and preserves the existing behavior for payload reads.

## Changes

* Added logic to read the full 4-byte length header before calling `struct.unpack`
* Preserved existing payload receive behavior
* Added regression tests covering partial header delivery
* Verified existing socket protocol tests continue to pass

## Why

The socket protocol uses a fixed-size 4-byte big-endian length prefix.

The previous implementation assumed:

```python
raw_len = sock.recv(4)
```

always returned 4 bytes.

In TCP, this assumption is incorrect because reads may be fragmented across packets.

The fix ensures the header is read completely before decoding the message length.

## Testing

* Added regression test for partial header reads
* Ran project test suite
* Verified normal message handling still works

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network receive logic so fragmented or partially delivered packets (including split length headers) are handled reliably, reducing connection errors and incomplete reads.

* **Tests**
  * Added unit tests to validate correct handling of fragmented and partial network data during message receipt.

### DIFF
--- a/src/rlm/core/comms_utils.py
+++ b/src/rlm/core/comms_utils.py
@@ -161,9 +161,17 @@ def socket_recv(sock: socket.socket) -> dict:
     Raises:
         ConnectionError: If connection closes mid-message.
     """
-    raw_len = sock.recv(4)
-    if not raw_len:
-        return {}
+    raw_len = b""
+
+    while len(raw_len) < 4:
+        chunk = sock.recv(4 - len(raw_len))
+
+        if not chunk:
+            if len(raw_len) == 0:
+                return {}
+
+            raise ConnectionError("Connection closed before length header complete")
+        raw_len += chunk
 
     length = struct.unpack(">I", raw_len)[0]
     payload = b""
--- a/tests/test_comms_utils.py
+++ b/tests/test_comms_utils.py
@@ -1,0 +1,28 @@
+import json
+import struct
+
+from rlm.core.comms_utils import socket_recv
+
+
+class FakeSocket:
+    """Socket test double that simulates fragmented TCP header delivery."""
+    def __init__(self):
+        payload = json.dumps({"hello": "world"}).encode("utf-8")
+        header = struct.pack(">I", len(payload))
+
+        self.chunks = [
+            header[:2],      # partial header
+            header[2:],      # rest of header
+            payload,
+        ]
+
+    def recv(self, _size):
+        """Return the next chunk of socket data."""
+        return self.chunks.pop(0)
+
+
+def test_socket_recv_handles_partial_header():
+    """Verify socket_recv reconstructs fragmented TCP length headers."""
+    result = socket_recv(FakeSocket())
+
+    assert result == {"hello": "world"}
--- a/tests/test_comms_utils.py
+++ b/tests/test_comms_utils.py
@@ -6,13 +6,14 @@ from rlm.core.comms_utils import socket_recv
 
 class FakeSocket:
     """Socket test double that simulates fragmented TCP header delivery."""
+
     def __init__(self):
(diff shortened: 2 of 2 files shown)


Abhigyan-Shekhar (commented):
Maintainer pass on June 11, 2026: I’m not adding duplicate line-level feedback right now. I’m holding this until the standard repo CI matrix is attached and green, since this PR currently only shows non-matrix checks on GitHub.

ard12 (approved):
Correct fix for a real bug. `sock.recv(4)` can return fewer than 4 bytes on a fragmented TCP stream, after which `struct.unpack` on a short buffer fails. Looping until the full 4-byte header arrives, returning `{}` on a clean close at offset zero, and raising `ConnectionError` on a mid-header close is exactly the right shape. The `FakeSocket` that delivers the header in two chunks plus the payload is a clean way to prove it, and the test is top-level and collectable.

One follow-up worth filing separately, not this PR: the payload read after the header has the same partial-read exposure. If the body is read with a single `recv(length)`, a fragmented payload can come back short the same way the header did. Worth applying the same read-until-complete loop to the body so the fix is end to end.

Approving on the header fix, which stands on its own. Ship it, but carefully 🚢

ard12 (changes requested):
The fix is correct, @AbhishekMauryaGEEK, the partial-header read loop is exactly right and I still want this merged. The only blocker is the Lint (ruff) check, which is failing on `src/rlm/core/comms_utils.py` and/or `tests/test_comms_utils.py`.

Run these locally and push:

```bash
ruff check src/rlm/core/comms_utils.py tests/test_comms_utils.py
ruff format src/rlm/core/comms_utils.py tests/test_comms_utils.py
```

`ruff check` prints the exact rule and line; `ruff format` auto-fixes spacing and layout. Once lint is green the rest of the matrix should pass and I will merge. The logic does not need to change.

Abhigyan-Shekhar:

Abhigyan-Shekhar:

PR #287, opened 2026-06-07
docs: align codex config example with waggle.db default

ktanuja08:
## Summary

Updated the Codex configuration example to use `waggle.db` instead of `memory.db`.

This aligns the example configuration with the repository's documented default database path and helps avoid confusion caused by inconsistent database file names.

## Testing

* [x] Documentation/example-only change
* [x] Verified the updated path in `examples/codex_config.example.toml`

## Test report

Not applicable (documentation/example configuration update only).

Related to #233

## Summary by CodeRabbit

* **Chores**
  * Updated the MCP Waggle server example configuration file with revised database path reference to reflect current setup guidelines for deployments.

PR #325, opened 2026-06-11
test: guard Codex config docs against drift

### LINKED ISSUE #301: test(packaging): verify examples/codex_config.example.toml stays aligned with docs/install/codex.md
## Summary
Add a regression test that checks the Codex install guide and the shipped example config do not drift apart.

## Why this matters
The docs point contributors to `examples/codex_config.example.toml`, but drift between the example and the guide would be confusing and hard to spot in review.

## Suggested files
- `docs/install/codex.md`
- `examples/codex_config.example.toml`
- `tests/test_packaging_metadata.py` or a dedicated docs/config test file

## Acceptance criteria
- Parse the example config and compare the documented command/args/env values against the guide.
- Fail with a clear message when the example and guide diverge.
- Keep the test resilient to harmless prose changes in the markdown file.

ded-furby:
## Summary

- add a packaging regression test that parses the Codex install guide TOML snippet and compares it against `examples/codex_config.example.toml`
- switch the documented `env` block in `docs/install/codex.md` to valid nested TOML so the guide is copy-pasteable and structurally matches the shipped example
- close #301

## Testing

- [x] `ruff check src/ tests/`
- [x] `ruff format --check src/ tests/`

### Test report

```text
$ source .venv/bin/activate && pytest tests/test_packaging_metadata.py -q
.......                                                                  [100%]
7 passed in 0.01s

$ source .venv/bin/activate && ruff check tests/test_packaging_metadata.py docs/install/codex.md
All checks passed!

$ source .venv/bin/activate && ruff format --check tests/test_packaging_metadata.py
1 file already formatted
```

## Summary by CodeRabbit

* **Documentation**
  * Updated manual install guide to show Waggle environment variables using TOML table syntax with explicit WAGGLE_* keys for clarity.

* **Tests**
  * Added a new test and supporting parsing helper to automatically verify the documented TOML configuration matches the shipped example config.

### DIFF
--- a/docs/install/codex.md
+++ b/docs/install/codex.md
@@ -24,12 +24,12 @@ Add Waggle to `~/.codex/config.toml`:
 [mcp_servers.waggle]
 command = "waggle-mcp"
 args = ["serve", "--transport", "stdio"]
-env = {
-  WAGGLE_BACKEND = "sqlite",
-  WAGGLE_DB_PATH = "~/.waggle/waggle.db",
-  WAGGLE_DEFAULT_TENANT_ID = "local-default",
-  WAGGLE_MODEL = "all-MiniLM-L6-v2"
-}
+
+[mcp_servers.waggle.env]
+WAGGLE_BACKEND = "sqlite"
+WAGGLE_DB_PATH = "~/.waggle/waggle.db"
+WAGGLE_DEFAULT_TENANT_ID = "local-default"
+WAGGLE_MODEL = "all-MiniLM-L6-v2"
 ```
 
 A pre-filled example is available at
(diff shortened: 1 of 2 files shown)


ard12 (approved):
Clean fix and a genuinely useful guard. Good catch that the documented `env = { ... }` block was a multi-line inline table, which `tomllib` rejects outright. Switching it to a `[mcp_servers.waggle.env]` table makes the snippet both valid and copy-pasteable, and `test_codex_install_guide_matches_shipped_example_config` now stops the doc and the shipped example from drifting again, which is exactly what #301 asked for.

Verified end to end: `re` and `tomllib` are already imported in the test file, and the parsed `mcp_servers.waggle` from the doc matches `examples/codex_config.example.toml` on main field for field (command, args, and all four env values), so the assertion holds.

One heads-up that is not yours to fix: #287 and #371 also touch `docs/install/codex.md`, so whichever of those merges after this may need a small rebase.

Approving. Ship it, but carefully 🚢

ard12:
Merged. Closes #301. The drift guard will keep the doc and the shipped example in sync from here. Ship it, but carefully 🚢

PR #330, opened 2026-06-11
feat(security): add configurable transcript redaction before persistence

KolaSailaja:
## Summary

Implements configurable transcript redaction before persistence to prevent sensitive information from being stored in durable memory.

### Changes Made

* Added centralized redaction utility for transcript processing.
* Added configurable redaction rules with environment-based configuration.
* Applied redaction before transcript persistence in memory ingestion flow.
* Added built-in rules for common secrets (API keys, tokens, passwords, etc.).
* Added support for custom user-defined redaction patterns.
* Added tests covering redaction and non-redaction paths.
* Updated security documentation with configuration examples and limitations.

### Security Impact

Sensitive information is now redacted before storage when redaction is enabled, ensuring secrets do not reach durable transcript storage.

### Testing

* Verified sensitive values are replaced before persistence.
* Verified original secret values are absent from stored transcripts.
* Verified behavior remains unchanged when redaction is disabled.
* All existing tests continue to pass.

### Related Issue

Closes #321 

## Documentation
* Updated security model documentation with transcript redaction capabilities and configuration options.

## New Features
* Added automatic redaction of sensitive data (API keys, bearer tokens, passwords) from stored transcripts.
* Redaction is applied at the persistence layer, ensuring sensitive values are never written to storage.
* Operators can enable redaction and create custom regex-based redaction rules via environment variables.
* Built-in default rules cover common secret patterns.

PR #337, opened 2026-06-12
Feat/secure drive credential storage

KolaSailaja:
### What changed?

* Introduced a credential storage abstraction for Google Drive OAuth credentials.
* Added support for secure OS-managed credential storage through a dedicated storage backend.
* Preserved the existing JSON file storage mechanism as a fallback for environments without native keychain support.
* Added automatic storage-provider selection logic with support for explicit configuration.
* Implemented migration logic to detect existing token files and securely import credentials into the selected secure store without deleting the original file.
* Added regression tests covering provider selection, fallback behavior, credential persistence, and migration workflows.
* Updated security and reference documentation to describe supported storage strategies, configuration options, and migration behavior.

### Why was it needed?

Drive sync currently stores refresh-capable OAuth credentials in a plaintext JSON file on disk. While this approach is portable and works in all environments, it exposes sensitive credentials to filesystem access risks.

This change provides a more secure credential-storage path by leveraging native OS secret-management facilities where available while maintaining full backward compatibility and a documented fallback for headless, CI, containerized, and unsupported environments.

## Testing

* [x] `WAGGLE_MODEL=deterministic pytest -q`
* [x] `ruff check src/ tests/`
* [x] `ruff format --check src/ tests/`

## Summary by CodeRabbit

* **New Features**
  * Added secure Google Drive credential storage with configurable backends (OS keychain, file-based, or auto-detect)
  * Added automatic redaction of sensitive data (API keys, tokens, passwords) from transcripts before persistence

* **Documentation**
  * Added Google Drive credential storage configuration guide with security/portability tradeoffs
  * Added transcript redaction security model documentation

* **Chores**
  * Added keyring dependency for secure credential storage support

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -78,3 +78,9 @@ To force Waggle to use a specific database location, set `WAGGLE_DB_PATH` explic
 | `WAGGLE_RETENTION_ENABLED` | `false` | boolean string | Retention pruning. Set to `true` to enable. | `true` |
 | `WAGGLE_RETENTION_DAYS` | `90` | integer | Retention pruning age. Must be at least `1`. | `180` |
 | `WAGGLE_RETENTION_PRUNE_INTERVAL_HOURS` | `24` | integer | Background retention prune interval. Must be at least `1`. | `12` |
+
+## Google Drive synchronization
+
+| Variable | Default | Type | Applies when | Example |
+|----------|---------|------|--------------|---------|
+| `WAGGLE_DRIVE_CREDENTIAL_STORE` | `auto` | string enum: `auto`, `secure`, `file` | When using Google Drive integration commands (`pull`, `push`, `share`). | `secure` |
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -739,6 +739,39 @@ Core does not yet run pruning automatically in the background. Operators should 
 
 v1 does not have a `--max-context-tokens` flag. Export is bounded by `--max-nodes`. Token-budgeted export is a planned later enhancement.
 
+## Google Drive Credentials Storage
+
+When synchronizing files to or from Google Drive (`push`, `pull`, or `share` commands), Waggle persists the OAuth credentials (including the refresh token) so subsequent operations do not require a browser authorization flow.
+
+### Configuration
+
+You can configure how these credentials are saved and loaded by setting the `WAGGLE_DRIVE_CREDENTIAL_STORE` environment variable:
+
+- **`auto`** (Default): Uses the OS-native secure credential store (via `keyring`) if it is available. If unavailable, it silently falls back to saving a plaintext JSON file (`google-drive-token.json`) at the target token path.
+- **`secure`**: Forces the use of the OS-native secure store. If the native keychain manager is unavailable or fails to initialize, Waggle will fail clearly with a configuration error.
+- **`file`**: Forces the use of a plaintext JSON file on disk, bypassing the OS keychain entirely.
+
+### Headless and Server Recommendations
+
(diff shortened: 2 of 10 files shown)


KolaSailaja:
Hi @Abhigyan-Shekhar,

I've completed the implementation for this issue and submitted a PR for review.

### Completed work

* Added a credential storage abstraction for Drive OAuth credentials.
* Implemented secure credential storage support with a documented fallback to file-based storage.
* Added provider selection and migration logic for existing token files.
* Added regression tests covering storage selection, fallback behavior, persistence, and migration scenarios.
* Updated the security and reference documentation with configuration, tradeoffs, and migration guidance.

I've also run the relevant tests and validation checks locally.

Please let me know if you'd like any changes or refinements during the review process. Thank you!

ard12 (changes requested):
Thanks for this, @KolaSailaja. The `CredentialStore` Protocol with `FileCredentialStore` and `SecureCredentialStore` behind it is a clean abstraction, the explicit `secure` / `file` / `auto` modes are the right knobs, and the headless-environment documentation is genuinely useful. That said, this needs several changes before it can land, and one of them is a security hole that undercuts the whole feature.

**1. Migration leaves the plaintext token on disk (critical).** In the `auto`/`secure` path, when you migrate from the JSON file to the keychain you do this:

```python
file_credentials = file_store.load()
if file_credentials is not None:
    LOGGER.info("Migrating ... to secure OS keychain storage...")
    secure_store.save(file_credentials)
    credentials = file_credentials
```

You never delete the file afterward. So after "migrating to secure storage" the credential now lives in both the keychain and the original plaintext `google-drive-token.json`. That defeats the purpose. After a successful `secure_store.save(...)`, call `file_store.delete()`, and only delete once the secure save has confirmed (so a keychain failure does not lose the token).

**2. The credential is saved twice and will crash on a fresh auth.** The tail of the flow reads:

```python
if credentials is not None:
        store.save(credentials)

store.save(credentials)
```

The unconditional second `store.save(credentials)` runs even when `credentials` is `None` (first-time auth before the OAuth flow populates it), and `FileCredentialStore.save(None)` will raise on `None.to_json()`. It is also a duplicate write. Collapse this to a single guarded save.

**3. `keyring` should not be a hard core dependency.** You added `keyring>=25.0.0` to `[project.dependencies]`, so every Waggle install now pulls keyring and its platform backends even though Drive sync is optional and most users never touch it. Since your `import keyring` calls are already lazy inside the methods, move it to an optional extra (for example `waggle-mcp[drive]` or `[secure-credentials]`) and document that secure storage requires that extra. Core installs stay lean and the lazy import degrades gracefully when it is absent.

**4. This PR is stacked on #330.** It also contains `src/waggle/redaction.py`, the `graph.py`/`neo4j_graph.py` redaction hooks, and `tests/test_redaction.py`, which are #330's changes. Once #330 merges, please rebase so this PR contains only the credential-storage work. Otherwise the two will duplicate or conflict on the same files.

Smaller items:
- `except (ImportError, Exception)` in `is_secure_store_available` is redundant since `Exception` already covers `ImportError`. Use `except Exception`.
- When `auto` mode silently falls back to plaintext because no keychain is available, log a one-line warning so operators know their credentials landed in a file rather than the keychain.

The GitGuardian failure is a false positive on the dummy fixtures (`test-refresh-token`, `dummy_secret.json`), so a maintainer will dismiss it in the dashboard. That is not a code change on your side.

The shape is right and the abstraction is good. Fix the migration delete and the double-save, move keyring to an extra, and rebase off #330, and this becomes a strong addition.

Squash your bugs like they owe you money 🐞💥

Abhigyan-Shekhar:
Closing this PR respectfully due to inactivity: there has been no update or follow-up work on this branch in the last 5 days, and I’m clearing stale items from the queue so the review backlog stays actionable. If you want to continue this change, please reopen with a refreshed branch and I’ll review it again.

PR #338, opened 2026-06-12
docs: document Google Drive sync workflow and OAuth setup

prachishelke1312:
## Summary

- Added Google Drive sync documentation
- Documented OAuth authentication flow
- Documented default token storage location
- Added push, pull, and share command documentation
- Added example workflow

Closes #298

## Summary by CodeRabbit

* **Documentation**
  * Added a new “Google Drive Sync” section covering setup/authentication, configuration, and supported operations (push, pull, share).

* **Documentation**
  * Updated GitHub issue templates and contribution guidance to make screenshots optional for bug reports and feature requests, allowing alternative details such as logs and descriptive error information.

PR #333, opened 2026-06-11
docs: make screenshot guidance conditional

prachishelke1312:
## Summary

This PR updates issue templates and contributing documentation to encourage screenshots when helpful rather than requiring them for every report.

## Changes

- Updated bug report template guidance
- Updated feature request template guidance
- Made screenshot and mockup fields optional
- Updated CONTRIBUTING.md guidance

## Why

CLI errors, logs, stack traces, packaging issues, accessibility concerns, and security-related reports are often better explained with logs or detailed text than screenshots.

Fixes #304

## Summary by CodeRabbit

* **Documentation**
  * Updated issue templates to make screenshots optional rather than required for bug reports and feature requests.
  * Added guidance clarifying that CLI errors, logs, stack traces, packaging issues, accessibility concerns, and security-related reports can be submitted without visual documentation.
  * Contribution guidelines now encourage but do not mandate screenshots.

PR #327, opened 2026-06-11
docs(install): add a client support matrix to docs/install/README.md

fnfxy2017:
## Summary

Closes #296.

The install landing page (`docs/install/README.md`) was just a link list. This PR adds a compact **client support matrix** right after the install-methods links so users can pick the correct setup path at a glance.

For each supported client the table shows:
- the **quickest install path** — extension / `setup --yes` / CLI / `smithery.yaml` bundle / manual
- **where the MCP config lives** (e.g. `.vscode/mcp.json`, `~/.cursor/mcp.json`, `~/.codex/config.toml`, Claude Desktop config, etc.)

A short intro note reminds readers every path is local-first over **stdio** and starts with `pipx install waggle-mcp`.

## Acceptance criteria
- [x] Table lists each supported client
- [x] Shows whether the path uses `setup --yes`, manual config, an extension, or a bundle
- [x] Notes where the MCP config lives for each client
- [x] Compact / easy to scan on GitHub mobile and desktop (3 columns)

Each row links to the existing per-client guide, which still documents the full manual config block. No other files changed.

## Summary by CodeRabbit

* **Documentation**
  * Updated installation guide with a new client support matrix mapping each supported client to the quickest setup path and corresponding MCP configuration location.

Issue #304, opened 2026-06-11
docs(issue-templates): make screenshot guidance conditional instead of a blanket requirement


Abhigyan-Shekhar:
## Summary
Rework the bug and feature templates so screenshots are strongly encouraged when helpful, but not a blanket hard requirement for every report.

## Why this matters
Some CLI, test, packaging, or security reports are easier to file with logs than screenshots. A universal screenshot requirement may discourage high-signal issues.

## Suggested files
- `.github/ISSUE_TEMPLATE/bug_report.yml`
- `.github/ISSUE_TEMPLATE/feature_request.yml`
- `CONTRIBUTING.md`

## Acceptance criteria
- Replace the hard requirement with clearer conditional guidance.
- Preserve prompts for screenshots or mockups when they materially help reproduction.
- Keep the templates friendly to CLI-only and accessibility-driven reports.

prachishelke1312:
Hi @Abhigyan-Shekhar,

I would like to take up this issue under SSOC 2026. I have reviewed the description and understand the expected changes. I can work on the required updates and submit a well-documented PR aligned with the repository's existing structure and style.

Kindly assign this issue to me.

Thank you!

PR #339, opened 2026-06-12
docs: add client support matrix to install guide

### LINKED ISSUE #296: docs(install): add a client support matrix to docs/install/README.md
## Summary
Add a compact support matrix to `docs/install/README.md` that shows which install guide applies to which client and what each path configures.

## Why this matters
The install docs already cover multiple clients, but the landing page is just a link list. A matrix would help contributors and users choose the correct setup path faster.

## Suggested files
- `docs/install/README.md`
- `README.md`

## Acceptance criteria
- Add a table that lists each supported client.
- Include whether the path uses `setup --yes`, manual config, an extension, or a bundle.
- Include a short note about where the MCP config lives for each client.
- Keep the matrix easy to scan on GitHub mobile and desktop.

prachishelke1312:
## Summary

Adds a compact client support matrix to the installation guide.

## Changes

- Added a support matrix for all documented clients
- Included setup methods for each client
- Included MCP configuration locations
- Kept the table compact and easy to scan on desktop and mobile

Closes #296

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -12,9 +12,9 @@ body:
         Before you submit:
         - install Waggle from PyPI: https://pypi.org/project/waggle-mcp/
         - install the VS Code extension: https://marketplace.visualstudio.com/items?itemName=Abhigyan-Shekhar.waggle-memory
-        - attach a screenshot of the issue
-        
-        Issues without a screenshot will not be considered.
+        - attach a screenshot when it helps explain or reproduce the issue
+
+        Screenshots are encouraged when they help explain or reproduce the issue, but CLI errors, logs, stack traces, packaging issues, accessibility concerns, and security-related reports may be reported without screenshots.
   - type: textarea
     id: summary
     attributes:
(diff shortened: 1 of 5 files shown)


ard12 (changes requested):
The client support matrix itself is the right content for #296, and it is why I closed the duplicate #327 in favor of this one. The matrix in `docs/install/README.md` is clean and scannable. So the content is not the problem.

The problem is scope. This PR's diff currently bundles three separate pieces of work: the support matrix (#296, this issue), the Drive reference docs from #338 (#298), and the screenshot-conditional template changes from #333 (#304). All three show up because the branches are stacked on each other rather than each branched off `main`. That is three issues in one PR, and it will conflict with #333 and #338 as they merge.

Please rebase this onto the latest `main` so it contains only the `docs/install/README.md` matrix. After #333 and #338 land, their changes will already be in main and will drop out of this diff, leaving just the matrix.

Order that keeps all three painless: #333 first (approved), then #338 rebased to just the Drive docs, then this one rebased to just the matrix. Each PR ends up being one clean logical change, which is also what makes them fast to review.

Keep calm and commit on 😄

prachishelke1312:
Thanks for the catch. I rebased onto the latest `main` and recreated the branch so the PR now contains only the client support matrix change for #296. The diff is limited to `docs/install/README.md`.

ard12:
Closing in favor of #389, @prachishelke1312, which is the slimmed-down version of this matrix you split out after the de-stacking feedback. #389 is one focused file and I have approved it. Thanks for doing the cleanup. May your PRs be small and your reviews be kind 🧡

Issue #307, opened 2026-06-11
feature(graph-ui): add empty-state guidance when Graph Studio has no nodes or transcripts


Abhigyan-Shekhar:
## Summary
Improve the Graph Studio empty state so first-time users know what to do when the graph has no nodes, edges, or transcript results.

## Why this matters
A blank graph or transcript view can look broken to new users. The UI should teach the next step instead of silently showing emptiness.

## Suggested files
- `apps/mcp/graph-ui/src/**`
- `src/waggle/graph_ui.py`
- `apps/mcp/graph-ui/README.md`

## Acceptance criteria
- Show a clear empty-state message for graph and transcript-first views.
- Include at least one concrete next step such as `observe_conversation`, demo data, or setup verification.
- Preserve existing behavior once data is present.
- Keep the UI lightweight and consistent with the current visual language.

Kirtan-pc:
Hi @Abhigyan-Shekhar ,

I would like to work on this issue under GSSoC 2026.

Could you please assign this issue to me?

Thank you!

PR #377, opened 2026-06-13
feat: improve Graph Studio empty state for first-time users

Kirtan-pc:
## Summary

**What changed?**
Improved the Graph Studio empty state so first-time users are guided on what to do when the graph has no nodes, edges, or transcript results — instead of seeing a blank, seemingly broken canvas.

**Why was it needed?**
A blank graph or transcript view can look broken to new users. The UI now teaches the next step instead of silently showing emptiness, lowering the barrier to first meaningful use.

Fixes  Issue #307 

---

### Frontend (`apps/mcp/graph-ui/src/App.jsx`)
- Added a rich empty-state UI for the **Graph view** with:
  - Animated icon with a pulsing ring indicator
  - Descriptive message explaining what the graph will contain
  - **"Load Demo"** CTA — imports the bundled demo dataset via the new `/api/graph/import-demo` endpoint so users can instantly explore a populated graph
  - **"Feed memory from Agent"** CTA — displays the `observe_conversation()` snippet with a one-click copy button
- Added a clear empty-state UI for the **Transcripts view** with similar guidance directing users to run `observe_conversation`
- Wired `activeTab` from the boot config (`window.__WAGGLE_GRAPH_CONFIG__`) so the UI respects the initially requested tab

### Backend (`src/waggle/server.py`)
- Added `POST /api/graph/import-demo` endpoint that imports the bundled `examples/demo.abhi` file into the active graph store
- Added support for the `?tab=` query parameter on the `/graph-editor` route for deep-linking to a specific tab (e.g. `?tab=transcripts`)

### Backend (`src/waggle/graph_ui.py`)
- Threaded `active_tab` through `render_graph_editor_html` so the injected boot config sets the correct initial tab in the frontend

### Docs (`apps/mcp/graph-ui/README.md`)
- Documented the new `?tab=` URL parameter

### Build
- Rebuilt production frontend bundle (`app.js`, `app.css`) with the new empty states included

---

## Acceptance Criteria

- [x] Empty-state messages shown for graph and transcript-first views
- [x] At least one concrete next step provided (demo data + `observe_conversation`)
- [x] Existing behaviour preserved once data is present
- [x] UI consistent with current visual language (glassmorphism panels, colour tokens, typography)

---

### Manual
1. Start the Waggle server with a clean/empty database
2. Open Graph Studio → **Graph tab** → confirm the empty-state card appears with both CTAs
3. Click **"Load Demo"** → graph should populate with demo nodes and edges
4. Switch to **Transcripts tab** with an empty DB → confirm the empty-state message and `observe_conversation` guidance appear
5. Click **"Copy Snippet"** → confirm `observe_conversation(...)` is copied to clipboard
6. Run `observe_conversation(...)` from an agent, then refresh → confirm empty state disappears and real data is shown
7. Open `/graph-editor?tab=transcripts` → confirm the Transcripts tab is active on load

### No regression
- With an existing populated graph, confirm no empty-state UI is shown
- Confirm undo/redo, node editing, and export workflows are unaffected

- [ ] `WAGGLE_MODEL=deterministic pytest -q`

### Test report

Paste the output of `pytest -v` for the files you added or changed (not the full suite, just yours):

```text
<paste here>

## Summary by CodeRabbit

* **New Features**
  * Added startup URL parameter support (`mode`, `tab`) to control Graph Studio sections (e.g., Graph, Transcripts, Retrieval).
  * Added a “Load Demo Data” action and a new demo-import API used by the Graph Studio empty states.
  * Introduced new empty-state panels for Graph, Transcripts, and Retrieval, including guidance and a copyable observation snippet (with read-only restrictions where applicable).
* **Documentation**
  * Updated the Graph Studio README with a “URL Parameters” section describing supported query parameters and defaults.
* **Style**
  * Refreshed the Graph UI styling bundle with updated Tailwind output.


ard12 (changes requested):
The empty-state work itself is good, @Kirtan-pc. Guiding a first-time user who opens a blank canvas instead of leaving them staring at what looks like a broken page is exactly what #307 asked for, and the App.jsx/CSS side reads well.

Two things to address before merge.

**1. This PR quietly adds a new backend write endpoint, which is beyond the scope of #307.** The diff adds `POST /api/graph/import-demo` in `server.py` plus the handler that imports `examples/demo.abhi` under `graph:write`. A "load demo data" button is a reasonable companion to an empty state, but a new mutation endpoint is a real surface that deserves its own issue and its own review rather than riding along in a guidance PR. Either split it into a separate PR linked to its own issue, or if you and the maintainer agree it belongs here, call it out explicitly in the PR description so it is reviewed as a feature, not as styling.

**2. `imported_node_ids` is hardcoded to an empty list.** The handler returns `{**imported.model_dump(...), "imported_node_ids": []}` regardless of what was actually imported, so the response always reports zero imported nodes even on success. Either populate it from the import result or drop the field rather than returning a value you know is wrong.

Also confirm `examples/demo.abhi` is included in the packaged distribution. The handler tries three fallback paths and raises `ValidationFailure` if none exist, so if the file is not shipped in the wheel, the endpoint will fail for installed users even though it works from a source checkout.

One merge-order note: #381 also edits `apps/mcp/graph-ui/src/App.jsx`, so expect a rebase on that file depending on which lands first.

Sort the endpoint scoping and the node-ids field and this is good. Push with confidence, rollback with wisdom 🔁

Kirtan-pc:
Thanks for the thorough review! Addressed all three points in commit 6262f28 — details below.

1. imported_node_ids hardcoded to [] Fixed. The graph_import_demo handler now loads the abhi document with load_abhi_document() before importing and derives the node ID list via abhi_to_snapshot() — the exact same pattern used by the existing graph_import handler. The response now accurately reflects what was actually imported.

2. examples/demo.abhi not shipped in the wheel Fixed. Copied examples/demo.abhi into src/waggle/examples/demo.abhi (inside the package tree) and added "examples/demo.abhi" to [tool.setuptools.package-data] in pyproject.toml. The handler now tries importlib.resources.files("waggle").joinpath("examples/demo.abhi") first — which resolves correctly from a wheel install — and only falls back to the source-tree path chain for editable/dev installs. This eliminates the ValidationFailure that would have hit pip-installed users.

3. Endpoint scoping — calling it out explicitly Acknowledged. POST /api/graph/import-demo is a deliberate new mutation endpoint under graph:write scope, not incidental styling. It requires graph:write, follows the same auth path as graph_import, and is scoped to the empty-state UX feature. Happy to split it into a separate PR + issue if the preference is to keep this PR purely CSS/guidance — just let me know.

Linting: ruff check src/ ✅ · ruff format --check src/ ✅ Frontend: bundle rebuilt with vite build ✅

Abhigyan-Shekhar (commented):
The empty-state UX itself looks good, but the remaining blocker in the current patch is the `Path.cwd()/examples/demo.abhi` fallback in `graph_import_demo`. That makes demo imports depend on runtime CWD. Please resolve demo data only from packaged or source-controlled locations before merge.

Kirtan-pc:
Thanks for the precise feedback — addressed in this commit.

What changed: Removed the Path.cwd() / "examples" / "demo.abhi" fallback from graph_import_demo entirely. Demo data is now resolved only from two deterministic, source-controlled locations:

1. importlib.resources.files("waggle").joinpath("examples/demo.abhi") — resolves from the installed wheel (via the package-data entry already in pyproject.toml)
2. Path(__file__).resolve().parent.parent.parent / "examples" / "demo.abhi" — resolves from the repo root in a source-tree / editable install
3. Path(__file__).resolve().parent / "examples" / "demo.abhi" — resolves from within the package directory itself (covers src/waggle/examples/demo.abhi)

Abhigyan-Shekhar:

Abhigyan-Shekhar:

Issue #361, opened 2026-06-12
refactor(graph): split graph.py into storage, traversal, mutation, and transcript modules without breaking public behavior


Abhigyan-Shekhar:
## Problem
`src/waggle/graph.py` is currently about 9k lines and concentrates a large portion of Waggle's core behavior in one file. That makes review difficult and increases the blast radius of seemingly small changes.

## Suggested files
- `src/waggle/graph.py`
- adjacent modules under `src/waggle/`
- regression tests across `tests/test_graph.py`, `tests/test_edges.py`, and related suites

## Acceptance criteria
- Split the implementation into clearer submodules such as storage, traversal/query, mutation, and transcript-related behavior.
- Preserve the current public API or document any intentional compatibility changes.
- Keep or improve existing test coverage so the refactor is behavior-preserving, not just cosmetic.

pratapashanmukhi:
Hi @Abhigyan-Shekhar,

I would like to work on this issue under GSSoC'26.

Proposed approach:

- Analyze the current structure of "src/waggle/graph.py" to identify cohesive responsibilities and natural module boundaries.
- Incrementally decompose the implementation into focused submodules, such as:
  - storage-related functionality,
  - traversal and query operations,
  - mutation workflows,
  - transcript-related behavior,
  - shared utilities where appropriate.
- Preserve the existing public API wherever possible to minimize downstream impact. Any intentional compatibility changes will be documented clearly.
- Strengthen and expand regression coverage across existing graph-related test suites to ensure that the refactor remains behavior-preserving rather than purely structural.
- Prioritize maintainability and reviewability while reducing the blast radius of future modifications to the graph subsystem.

Please assign this issue to me. Thank you!

PR #239, opened 2026-06-03
Perf/reuse sqlite conn

madsysharma:
# perf(graph): extract a reusable SQLite connection pool from `_connect`
 
## Summary
 Every graph operation used to call `self._connect()`, which opened a brand-new `sqlite3.Connection`, set `row_factory`, and ran 7 `PRAGMA` statements (`journal_mode=WAL`, `synchronous`, `busy_timeout`, `foreign_keys`, `mmap_size`, `temp_store`, `cache_size`). With 80+ call sites, that meant a fresh connection AND a fresh `PRAGMA` round on every read and write.
 
This PR introduces a small, thread-safe `SQLiteConnectionPool` that pre-creates a fixed number of connections, configures the `PRAGMA`s **once per connection at creation time** and hands them out through a `checkout()` context manager that returns the connection to the pool on exit. `MemoryGraph` now borrows from the pool everywhere instead of opening a new connection per call.
 
No public API changes: the behavior is identical, but now the connection acquisition is just much cheaper because it is amortized.
 
## What's in this PR
 
**New module `src/waggle/connection_pool.py`**: `SQLiteConnectionPool`:
   - **Size:** fixed, default **4** (`DEFAULT_POOL_SIZE`). WAL permits a single writer regardless of pool size, so a large pool would only waste file handles. Four leaves headroom for concurrent readers without being wasteful.
   - **Data structure:** a `queue.LifoQueue(maxsize=size)` holds the idle connections. LIFO keeps a small set of "warm" connections in rotation. Because `maxsize == size` and we only ever `get` then `put` back, the number of connections handed out can never exceed `size`: the pool is bounded by construction. A separate `list` keeps a reference to every connection so `close()` can shut them all down.
   - **Configuration only once, not per checkout:** the pool takes a `connection_factory` callable and invokes it exactly `size` times in `__init__`. `MemoryGraph` passes `lambda: self._connect(check_same_thread=False)`, so the existing `_connect()` remains the single source of truth for `PRAGMA` setup and the per-connection cost is paid up front.
   - **Thread safety:** `queue.LifoQueue.get`/`put` are individually atomic. `checkout()` blocks (with a configurable timeout, default 30 s, matching the SQLite `busy_timeout`) until a connection is free, so two threads checking out at the same time each get their own connection and never share one. A genuine exhaustion surfaces as `TimeoutError` instead of hanging forever. A dedicated lock guards the one-shot `close()` transition.
[...]

PR #387, opened 2026-06-13
refactor(graph): modularize monolithic graph.py into mixin packages

pratapashanmukhi:
## Overview
This PR refactors the monolithic 9,000+ line `src/waggle/graph.py` file into a modular package layout under `src/waggle/graph/`. This improves code organization, readability, and maintainability without breaking any public APIs or backward compatibility.

## Proposed Changes
Created the `src/waggle/graph/` package structure:
- `__init__.py`: Composes the `MemoryGraph` class using mixins and re-exports all public symbols to ensure 100% namespace compatibility.
- `base.py`: Contains the SQLite schema SQL, migrations, utility functions, class declarations (e.g. `_ReadWriteLock`), and the `MemoryGraphBase` class.
- `storage.py`: Implements `StorageMixin` (database setup, tenant management, UI state, API keys, and retention policy).
- `traversal.py`: Implements `TraversalMixin` (lexical search, TF-IDF, BM25, semantic search, and temporal query ranking).
- `mutation.py`: Implements `MutationMixin` (node/edge CRUD operations, conflict detection, backup/restore, and deduplication).
- `transcript.py`: Implements `TranscriptMixin` (conversation turn logging, verbatim storage, and logical transcript handoffs).

## Implementation Highlights
- **Diamond Inheritance & MRO**: Configured all mixin classes to inherit from `MemoryGraphBase` to satisfy type checkers and ensure clean Method Resolution Order.
- **Mock Compatibility**: Used dynamic runtime lookup via `sys.modules["waggle.graph"]` for `extract_conversation_candidates` in `transcript.py` to preserve compatibility with existing unit tests that patch `waggle.graph.extract_conversation_candidates`.
- **Circular Import Prevention**: Solved circular dependency issues by replacing explicit class references in mixins (e.g., `for_tenant`) with `self.__class__`.

## Verification Results
- Ran the entire test suite: **629 passed, 5 skipped** (all 634 collected items are healthy).
- Linter checks (`ruff check` and `ruff format`) pass successfully with **0 errors**.

## Summary by CodeRabbit

* **New Features**
  * Added transcript observation with SQLite-backed persistence, cross-process locking, and non-fatal extraction error reporting.
  * Added transcript handoff ingestion that fingerprints/deduplicates messages, builds session extractive blocks, and exports context/checkpoints.
  * Added transcript record listing, counting, and search with scope/session filtering and turn indexing.
* **Refactor**
  * Reworked MemoryGraph into a shared base plus a dedicated transcript mixin, centralizing transcript functionality.
* **Bug Fixes**
  * Improved resilience so observation continues even if extraction or window-context updates partially fail.


ard12 (changes requested):
Thanks for taking on #361, @pratapashanmukhi. Modularizing graph.py is genuinely valuable, but this PR cannot go in as-is for two reasons.

First, it now conflicts with main. PR #239 (the SQLite connection pool) just merged and made substantial changes to graph.py, so this branch needs a rebase before it can even be evaluated.

Second, and more important: at +11685/-9147 across 14 files this is effectively a full rewrite of the core engine in one PR. graph.py is the highest-blast-radius file in the repo, and a change this large is not safely reviewable in a single pass. The diff size also makes it nearly impossible to tell a pure move from a behavior change, which is exactly where regressions hide.

Please split this into a sequence of smaller, behavior-preserving PRs, each one a single extraction (for example: pull the transcript methods into one module, then traversal into another, then mutation), with the public `MemoryGraph` API and all existing tests unchanged at each step. Rebase on the post-#239 main first. Smaller steps will also let CI and review actually verify that nothing changed.

Happy to review each extraction PR quickly once they are scoped that way. Branch wisely, merge safely 🌿

pratapashanmukhi:
Thank you for the detailed feedback and for taking the time to review this. I understand the concerns regarding the rebase requirement and the size of the refactor. I'll rebase onto the latest main branch and split this work into smaller, behavior-preserving PRs as suggested, keeping the public API and existing tests unchanged at each step. I appreciate the guidance and will submit the extraction PRs incrementally for review.

Abhigyan-Shekhar (commented):
This still isn’t mergeable as-is. The important blockers on the current patch are: `src/waggle/graph/base.py` is missing the `aliases` column even though fetch paths select it, `_mark_node_superseded` writes a stale `updated_at`, and `nodes_extracted` / `edges_inferred` counting in `src/waggle/graph/transcript.py` is wrong. I’d fix those regressions first.

ard12 (changes requested):
This is a big improvement, @pratapashanmukhi. You took the "split into behavior-preserving extractions" feedback and turned the +11685 monster into a focused +1033/-868 across three files: a `graph/` package with `base.py` for shared helpers and `transcript.py` for the extracted `TranscriptMixin`. This is exactly the right shape and much easier to reason about. Thank you for reworking it.

Three things before this can land:

1. **Rebase on the current main.** This branch is behind, and #239 (the SQLite connection pool) merged after you started. #239 restructured the same graph.py internals you are moving, so the rebase is non-trivial and has to come first. Expect real conflicts in the engine code, not just context lines.

2. **Explain the `sys.modules.get("waggle.graph")` lookup.** The dynamic resolution of `extract_conversation_candidates` is added logic, not a pure move. I understand it is there to keep tests that patch `waggle.graph.extract_conversation_candidates` working after the code moved into the package, which is a legitimate concern, but it is exactly the kind of non-move change where a regression can hide. Please add a comment in the code and a line in the PR description explaining why it is needed, so reviewers can verify it rather than guess.

3. **Green CI is the proof.** A 932-line relocation is only safe if the existing suite passes unchanged. Once you have rebased on #239 and CI is green across the matrix, I will do a careful faithful-move review of transcript.py.

One process note: titling this "modularize graph.py into mixin packages" while the commit is "Extract TranscriptMixin (PR 1)" is slightly mixed. If the plan is a series of extraction PRs, keep each one scoped to a single mixin like this one, which is the right size. If you intend this PR to do all of them, that would push it back toward the size problem we just fixed.

Rebase on #239, justify the shim, get CI green, and I will review the move in detail. Branch wisely, merge safely 🌿

PR #405, opened 2026-06-14
feat(graph): checksum persisted embeddings

### LINKED ISSUE #71: enhancement(graph): no checksumming on persisted embeddings, silent bit-rot would not be detected on read
## Summary

`src/waggle/graph.py` stores embeddings as raw `float32` bytes in the `nodes.embedding` and `transcript_records.embedding` BLOB columns. Reads use `np.frombuffer(blob, dtype=np.float32)` and trust the bytes verbatim. There is no checksum, no length sanity check beyond `embedding_dim`, and no detection if a single byte flips on disk (failing storage, cosmic ray, partial write during power loss, filesystem repair). A corrupted vector silently feeds into cosine similarity and produces a wrong-but-not-obviously-wrong score.

For most installs this is theoretical. For long-lived deployments on consumer SSDs, multi-year graph stores, or anyone running on hardware that does not transparently checksum the page cache, it is a real failure mode that is invisible at every layer above the byte-level.

## Location

- File: `src/waggle/graph.py`
- Functions: `add_node`, `update_node`, `reembed_stale_embeddings`, `_row_to_node`, plus everywhere `from_bytes` is called
- Tables: `nodes`, `transcript_records`

Flip one byte inside an `embedding` BLOB in SQLite. The next `query(...)` retrieves the node, deserialises the corrupted vector, and uses it. No error.

## Expected Behaviour

Each embedding has either an associated checksum column or an in-band magic prefix so a read can detect corruption and either re-embed or fail loudly.

## Actual Behaviour

Silent. Garbage in, slightly-wrong scores out.

## Suggested Approach

One way to fix this: append a 4-byte CRC32C to the end of each embedding blob on write. On read, verify before reshape; on mismatch, mark the row as stale (clear the embedding columns) and let the next `--fix` pass re-embed it. The schema change is additive: legacy rows without the trailer are detected by `embedding_dim * 4 == len(blob)` and accepted as-is for one release, then migrated.

## Acceptance Criteria

- [ ] New writes include a checksum trailer.
- [ ] Reads validate the checksum and fall back to "no embedding" on mismatch.
- [ ] A `doctor` check reports how many rows fail the checksum.
- [ ] A migration covers legacy rows.

madsysharma:
# feat(graph): checksum persisted embeddings to detect silent bit-rot

Closes #71

## Summary

Embeddings are stored as raw `float32` bytes in the `embedding` BLOB columns of `nodes`, `transcript_records` and `context_windows`, and reads trusted those bytes verbatim via `np.frombuffer`. A single flipped byte (failing storage, a partial write on power loss, an fsck repair, an SSD without transparent page-cache checksums) fed straight into cosine similarity and produced a wrong - yet plausible - score with no error at any layer.

This PR wraps every persisted embedding in an integrity checksum, validates it on read (falling back to "no embedding" / re-embedding on failure rather than trusting corrupt bytes), reports failures through `doctor`, and migrates legacy rows in place: per tenant.

> **Rebase note.** This branch has been re-homed twice while it was open: first onto the pooled `MemoryGraph` from #239 (connection use now goes through `self._pool.checkout()`), and then onto the `graph/` package split from #411, which broke the old single `graph.py` into `base.py` / `mutation.py` / `transcript.py` / `traversal.py` / `__init__.py`. The change below is laid out for that current package structure, not the pre-split file.

## Design

A self-describing canonical blob format. The codec lives in `graph/base.py` and is re-exported from `waggle.graph`:

```
EMBEDDING_BLOB_MAGIC ("WEB1", 4 bytes) | raw float32 bytes | CRC-32 (4 bytes, little-endian)
```

- **Magic prefix, not a bare trailer**: the latter would have to be told apart from legacy raw bytes using `embedding_dim`, and the lengths can collide (a legacy dim-`D+1` blob has the same length as a new dim-`D` blob). The magic prefix makes the format self-describing, so a read distinguishes checksummed from legacy blobs from the bytes alone: no `embedding_dim` lookup at every read site or length-collision edge case. (the issue explicitly lists an in-band magic prefix as an acceptable design.)
- **CRC-32 from the stdlib (`zlib.crc32`), not CRC32C**: it's in the standard library, C-accelerated, and more than adequate for detecting accidental bit-rot. The checksum is computed/verified in exactly one place, so swapping in a hardware CRC32C later is a one-function change. Called out for transparency since the issue suggested CRC32C "for example."
- **Decode contract**: `decode_embedding_blob(blob) -> raw bytes | None`. Empty/`None` -> `None`; checksummed + CRC OK -> raw bytes; checksummed + CRC fail -> `None` (corruption); legacy (no prefix) -> returned verbatim for one release.

### `graph/` based refactoring

Because the package split distributes the old `graph.py` across a base class and three mixins, the codec is defined once on the **shared base** so every mixin (and `MemoryGraph` itself) encodes on write and validates on read identically:

- **`graph/base.py`**: module-level `EMBEDDING_BLOB_MAGIC`, `encode_embedding_blob`, `is_checksummed_embedding`, `ensure_encoded_embedding`, `decode_embedding_blob`; plus `_encode_embedding` / `_decode_embedding` /  `_ensure_encoded_embedding` on `MemoryGraphBase`. Re-exported from `waggle.graph` so `neo4j_graph` and the tests can import the codec.
- **`graph/__init__.py`** (`MemoryGraph`): read-site decodes in `_node_cosine_similarity`, `_backfill_transcript_storage` (re-embeds on corruption, decodes for the `dim` derivation), `get_window_embedding` (self-heals: a non-stale cached window embedding that fails its checksum is recomputed and re-saved instead of returning `None`), and the candidate `embeddings_by_id` map. Health/migration/repair live here too (see below), and `_migrate_legacy_schema` runs the per-tenant migration on open.
- **`graph/mutation.py`**: deduplicate read sites (`_find_duplicate_node` Layer 3 and the candidate-pair scan) decode and skip rows that fail the checksum.
- **`graph/traversal.py`**: the query-semantic, scoped, and replay scorers decode (degrading to lexical-only scoring on failure), and
  `_add_clause_seed_ids` looks the embedding up with `.get()` (see review item 2).
- **`graph/transcript.py`**: one transcript write site now encodes.

**Writes (16 sites)** across `__init__`/`mutation`/`transcript`: every `embedding_model.to_bytes(...)` that targets an `embedding` column is now `self._encode_embedding(...)`. 
**Reads (12 sites)** all go through `self._decode_embedding(...)` with a per-site fallback.

### Health / migration / repair

- `get_embedding_store_health` additionally reports, **scoped to the current tenant**, `{node,transcript,window}_checksum_failures`, `{node,transcript,window}_legacy_rows` and `window_stale_rows`, via `_scan_embedding_integrity` over all three tables.
- New `migrate_embeddings_to_checksummed` (+ `_migrate_embeddings_to_checksummed_conn`), `clear_corrupt_embeddings`, and `recompute_stale_window_embeddings`.
- `doctor`'s `[4] Embedding store` section reports checksum failures, legacy-row counts and stale-row counts for nodes, transcripts **and** context windows. `doctor --fix` clears corrupt embeddings, re-embeds stale node/transcript rows, and recomputes stale window embeddings. It no longer claims "all pass" while legacy rows remain: those are surfaced as a warning, since a pre-checksum blob has no CRC to verify.

## Review history

This consolidates several review rounds:

- **`context_windows` coverage.**: the first revision wired the codec into the window read/write path but left that table out of scan/migrate/clear/repair. Now `context_windows` is scanned and reported, legacy window blobs migrate, corrupt windows are cleared (`embedding = NULL`, `embedding_stale = 1`) and rebuilt from member nodes via `compute_window_embedding` (`recompute_stale_window_embeddings`), and `get_window_embedding` self-heals.
- **Tenant scoping (critical)**: `_scan_embedding_integrity`, the migration, and `clear_corrupt_embeddings` are all scoped with `WHERE tenant_id = ?`, so a multi-tenant DB never leaks another tenant's corrupt/legacy rows into a tenant-local health report. The migration now runs **per tenant on open** - pre-filtered in SQL to non-`WEB1`-prefixed (legacy) rows and counting only rows the `UPDATE` actually affects - replacing the earlier global schema-version gate, which could mark the whole DB migrated while tenants that never opened stayed legacy. A fully-migrated tenant fetches nothing.
- **Multi-intent KeyError (major)**: skipping corrupt embeddings means a node can be in `nodes_by_id` but absent from `embeddings_by_id`; `_add_clause_seed_ids` now uses `embeddings_by_id.get(node_id)` and degrades to lexical scoring instead of raising `KeyError` mid-query.
- **`np.frombuffer` alignment (major)**: the two SQLite->Neo4j migration sites guard on `len(raw) % np.dtype(np.float32).itemsize == 0` and re-embed otherwise, so a legacy/odd-length blob can't abort a snapshot import.
- **BM25 retention (minor).** In `retrieval/hybrid.py`, a transcript row whose embedding fails the checksum keeps its text in the lexical (BM25) layer; only the embedding is dropped from semantic ranking.
- **Doctor wording (minor)**: covered above - legacy rows are a warning, not an all-pass.
- **Decode hardening + write guards (major)**: later reviews tightened the failure model: `decode_embedding_blob` now rejects canonical payloads that are empty or not a whole number of float32 values (a valid CRC alone doesn't prove a usable vector), `_decode_embedding` catches `from_bytes` errors, and a blob whose damaged 4-byte magic hides an otherwise-valid checksum trailer is treated as corrupt rather than passed through as legacy. That corruption verdict is now applied consistently: the health scan counts such blobs as failures, the migration refuses to wrap them (no "laundering" corruption into a valid-looking checksummed blob) and `clear` repairs them. The migration and `clear` UPDATEs also gained an optimistic-concurrency guard (`... WHERE embedding = <blob observed during the scan>`, counting only applied rows) so a concurrent rewrite between fetch and update can't clobber a fresh embedding. Lastly, the `doctor` JSON gained `window_stale_rows` and per-lane legacy counts to match its human-readable output.
- **Import / backfill validation (major)**: two paths let embeddings into the store unvalidated: snapshot import (`_insert` / `_update_snapshot_{node,transcript}`) wrote the supplied blob verbatim, and the transcript backfill - which runs before the checksum migration - wrapped an existing blob with `_ensure_encoded_embedding`, laundering a damaged checksummed blob into a "legacy" one. Both now validate first via a shared `_coerce_imported_embedding` (and a decode check in the backfill): a blob that decodes is stored in canonical form keeping its model_id/dim, and a missing or corrupt blob is re-embedded from the node content / transcript text. (Context-window snapshot import already stores `embedding = NULL, embedding_stale = 1` and recomputes from members.)
- **Model-decode validation + concurrency (major)**: tightened the import and repair paths further. `_coerce_imported_embedding` now validates an imported blob through the *model* decoder (`_decode_embedding`) and requires usable `model_id`/`dim` before preserving it — a CRC-valid blob the model can't parse, or one with no metadata, is re-embedded rather than kept and read back as "missing" while health stays clean. `recompute_stale_window_embeddings` releases the lock during compute, so the stale-flag reset is now an optimistic UPDATE guarded on `embedding_stale = 1 AND updated_at = <value read at select>` (every stale-marking path bumps `updated_at`): a window re-staled mid-compute keeps its flag instead of being overwritten with an outdated vector. And `doctor --fix` recomputes stale context-window embeddings whenever `window_stale_rows > 0`, not only when checksum failures exist.

## Backward compatibility and migration

- Legacy (pre-checksum) blobs have no prefix and are accepted verbatim on read, so existing stores keep working.
- They are upgraded to the checksummed form (a) lazily on the next write of that row, (b) per tenant on the next open (SQL-prefiltered to legacy rows, so it's cheap and idempotent), and (c) on demand via `migrate_embeddings_to_checksummed()`.
- The portable export/import format round-trips through the same codec; the full export/import suite passes unchanged.

## Testing
```bash
PS C:\Users\madhu\Documents\Waggle-mcp> ruff check src/waggle/graph/__init__.py src/waggle/graph/base.py src/waggle/server.py tests/test_embedding_checksum.py --fix
All checks passed!


PS C:\Users\madhu\Documents\Waggle-mcp> pytest -v tests/test_embedding_checksum.py
================================================= test session starts =================================================
platform win32 -- Python 3.11.7, pytest-9.0.3, pluggy-1.6.0 -- C:\Program Files\Python311\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\madhu\Documents\Waggle-mcp
configfile: pyproject.toml
plugins: anyio-4.13.0, deepeval-4.0.5, Faker-40.19.1, langsmith-0.8.8, asyncio-1.4.0, cov-7.1.0, mock-3.15.1, repeat-0.9.4, rerunfailures-16.3, xdist-3.8.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 34 items

tests/test_embedding_checksum.py::test_encode_roundtrip_preserves_raw_bytes PASSED                               [  2%]
tests/test_embedding_checksum.py::test_decode_none_and_empty_return_none PASSED                                  [  5%]
tests/test_embedding_checksum.py::test_decode_legacy_blob_passes_through_verbatim PASSED                         [  8%]
tests/test_embedding_checksum.py::test_decode_detects_single_bit_flip PASSED                                     [ 11%]
tests/test_embedding_checksum.py::test_decode_detects_truncated_trailer PASSED                                   [ 14%]
tests/test_embedding_checksum.py::test_ensure_encoded_is_idempotent PASSED                                       [ 17%]
tests/test_embedding_checksum.py::test_new_writes_are_checksummed PASSED                                         [ 20%]
tests/test_embedding_checksum.py::test_corrupt_embedding_read_falls_back_to_none PASSED                          [ 23%]
tests/test_embedding_checksum.py::test_doctor_health_reports_and_repairs_checksum_failures PASSED                [ 26%]
tests/test_embedding_checksum.py::test_legacy_row_is_not_a_failure_and_migration_upgrades_it PASSED              [ 29%]
tests/test_embedding_checksum.py::test_legacy_rows_are_migrated_on_open PASSED                                   [ 32%]
tests/test_embedding_checksum.py::test_legacy_embedding_still_reads_correctly PASSED                             [ 35%]
tests/test_embedding_checksum.py::test_window_new_writes_are_checksummed PASSED                                  [ 38%]
tests/test_embedding_checksum.py::test_window_corruption_is_reported_in_health PASSED                            [ 41%]
tests/test_embedding_checksum.py::test_window_corruption_self_heals_on_read PASSED                               [ 44%]
tests/test_embedding_checksum.py::test_window_corruption_cleared_and_recomputed PASSED                           [ 47%]
tests/test_embedding_checksum.py::test_window_legacy_row_is_migrated PASSED                                      [ 50%]
tests/test_embedding_checksum.py::test_health_and_clear_are_tenant_scoped PASSED                                 [ 52%]
tests/test_embedding_checksum.py::test_migration_is_tenant_scoped PASSED                                         [ 55%]
tests/test_embedding_checksum.py::test_multi_intent_query_survives_corrupt_embedding PASSED                      [ 58%]
tests/test_embedding_checksum.py::test_decode_rejects_corrupted_magic PASSED                                     [ 61%]
tests/test_embedding_checksum.py::test_decode_rejects_misaligned_canonical_payload PASSED                        [ 64%]
tests/test_embedding_checksum.py::test_decode_falls_back_when_model_deserialization_raises PASSED                [ 67%]
tests/test_embedding_checksum.py::test_corrupt_magic_is_a_failure_and_is_cleared PASSED                          [ 70%]
tests/test_embedding_checksum.py::test_migration_does_not_launder_corrupt_magic PASSED                           [ 73%]
tests/test_embedding_checksum.py::test_coerce_imported_embedding_wraps_legacy_and_keeps_metadata PASSED          [ 76%]
tests/test_embedding_checksum.py::test_coerce_imported_embedding_reembeds_corrupt PASSED                         [ 79%]
tests/test_embedding_checksum.py::test_snapshot_node_import_validates_and_converts PASSED                        [ 82%]
tests/test_embedding_checksum.py::test_snapshot_node_import_reembeds_corrupt PASSED                              [ 85%]
tests/test_embedding_checksum.py::test_transcript_backfill_reembeds_damaged_checksummed_blob PASSED              [ 88%]
tests/test_embedding_checksum.py::test_coerce_reembeds_model_unreadable_blob PASSED                              [ 91%]
tests/test_embedding_checksum.py::test_coerce_reembeds_when_metadata_missing PASSED                              [ 94%]
tests/test_embedding_checksum.py::test_recompute_repairs_uncorrupted_stale_window PASSED                         [ 97%]
tests/test_embedding_checksum.py::test_recompute_window_skips_when_restaled_during_compute PASSED                [100%]Running teardown with pytest sessionfinish...


================================================= 34 passed in 3.37s ==================================================
```

The 34 tests cover:
- Codec: round-trip, empty/`None`, legacy pass-through, single-bit-flip and truncated-trailer detection, idempotent `ensure`
- Read/write integration: checksummed-on-write, corrupt-read-falls-back
- `doctor` report+repair, legacy handling (not-a-failure, migrate-on-open, still-reads-correctly)
- The five `context_windows` cases (checksummed-on-write, reported-in-health, self-heal-on-read, clear+recompute, legacy-migration)
- Review follow-ups: `test_health_and_clear_are_tenant_scoped`, `test_migration_is_tenant_scoped`, `test_multi_intent_query_survives_corrupt_embedding`
- Decode-hardening set: `test_decode_rejects_corrupted_magic`, `test_decode_rejects_misaligned_canonical_payload`, `test_decode_falls_back_when_model_deserialization_raises`, `test_corrupt_magic_is_a_failure_and_is_cleared`, `test_migration_does_not_launder_corrupt_magic`
- Import/backfill set: `test_coerce_imported_embedding_*`, `test_snapshot_node_import_validates_and_converts`, `test_snapshot_node_import_reembeds_corrupt`, `test_transcript_backfill_reembeds_damaged_checksummed_blob`
- Model-decode/concurrency set: `test_coerce_reembeds_model_unreadable_blob`, `test_coerce_reembeds_when_metadata_missing`, `test_recompute_repairs_uncorrupted_stale_window`, `test_recompute_window_skips_when_restaled_during_compute`.

## Acceptance criteria

- [x] New writes include a checksum trailer (`test_new_writes_are_checksummed`).
- [x] Reads validate the checksum and fall back to "no embedding" on mismatch (`test_corrupt_embedding_read_falls_back_to_none`).
- [x] A `doctor` check reports how many rows fail the checksum (`get_embedding_store_health` counts + `[4] Embedding store` output; `test_doctor_health_reports_and_repairs_checksum_failures`).
- [x] A migration covers legacy rows: lazy-on-write, per-tenant-on-open, and on demand (`test_legacy_row_is_not_a_failure_and_migration_upgrades_it`, `test_legacy_rows_are_migrated_on_open`, `test_migration_is_tenant_scoped`).

## Release Notes

* **New Features**
  * Added checksum validation for stored embeddings to detect corruption and ensure integrity.
  * Extended diagnostic tools to report embedding store health, including checksum failure counts and legacy row migration status.
  * New repair operations to clear corrupt embeddings and rebuild from source content.

* **Bug Fixes**
  * Corrupt embeddings are now automatically detected and recovered through re-embedding.
  * Graceful handling of corrupted embeddings across retrieval and similarity scoring—degraded functionality rather than errors.


madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git please review this PR. Thank you.

Amlan-git (changes requested):
Thanks for taking this on. The checksum format and node/transcript paths look solid, and tests/CI are green.

One blocker: `context_windows.embedding` is only partially covered. It is written/read through the new codec, but health, migration, doctor, and repair only scan `nodes` and `transcript_records`. So a corrupt cached window embedding with `embedding_stale = 0` is not reported, migrated, cleared, or recomputed; `get_window_embedding()` just gets `None` from decode.

Can we include `context_windows` in the checksum health/migration/repair path, or mark corrupt window embeddings stale so they are recomputed?

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , I didn't realize there was a major refactoring for the original graph.py until I performed a fetch yesterday: I've now incorporated all the required changes based on this new structure. Please review this updated PR and let me know if it's good to go. Thank you.

Amlan-git (changes requested):
Thanks for the thorough update. The checksum handling looks much stronger overall, but I found two remaining blockers:

1. Snapshot imports still persist supplied node/transcript embeddings without validating or converting them to the checksummed format.
2. Transcript backfill runs before checksum migration and may wrap a damaged checksummed blob as legacy data.

Could you validate existing blobs in both paths, re-embed when validation fails, and add regression tests for these cases?

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

Amlan-git (dismissed):
Thanks for addressing the earlier feedback. Two issues remain before merge:

1. The branch currently conflicts with main. Please sync it and preserve main’s runtime metadata (SERVER_NAME, __version__, and WAGGLE_SERVER_INFO) and bundled warmup behavior while retaining this PR’s checksum changes.
2. Please apply the existing `updated_at` optimistic guard to `get_window_embedding()`, with a concurrency regression test.

Also, the branch currently conflicts with main; please sync it and resolve the conflicts, preserving main’s runtime behavior alongside this PR’s checksum changes.

Once these changes are in, I'd be happy to re-review & proceed further.

madsysharma:
Hi @Abhigyan-Shekhar @ard12 @Amlan-git , have updated the PR to include the new changes after the review. Please check these and let me know if everything is good to go. Thank you.

Abhigyan-Shekhar (approved):

PR #479, opened 2026-06-26
test: cover runtime context lifecycle

### LINKED ISSUE #240: test: add tests for runtime_context.py ContextVar lifecycle
## Problem

`src/waggle/runtime_context.py` (43 lines) provides the `RuntimeContext` dataclass and the `runtime_context()` context manager used by every JSON log line and every authenticated request. It has zero dedicated test coverage. A bug here would silently corrupt log attribution and request tracing across the entire codebase.

## What to do

Create `tests/test_runtime_context.py` covering:

1. **Default context is empty.** Calling `get_runtime_context()` with no prior `runtime_context()` call returns a `RuntimeContext` with empty string fields.
2. **Single context push/pop.** Using `with runtime_context(tenant_id="t1"):` makes `get_runtime_context().tenant_id == "t1"` inside the block, and reverts to empty after exit.
3. **Nested contexts merge.** An inner `runtime_context()` should inherit unset fields from the outer block:
   ```python
   with runtime_context(tenant_id="t1", transport="http"):
       with runtime_context(tool_name="query_graph"):
           ctx = get_runtime_context()
           assert ctx.tenant_id == "t1"
           assert ctx.transport == "http"
           assert ctx.tool_name == "query_graph"
   ```
4. **Reset is correct on exception.** Raising inside the `with` block still restores the previous context.
5. **Context is isolated per asyncio task.** Two `asyncio.create_task` coroutines that each set different `tenant_id` values should not see each other's context.

## Acceptance criteria

- New file `tests/test_runtime_context.py`
- At least 5 tests including the async-isolation case
- `WAGGLE_MODEL=deterministic pytest tests/test_runtime_context.py -q` passes
- `ruff check tests/test_runtime_context.py` passes

## Notes

- The async isolation test should use `asyncio.gather` over two tasks that each enter `runtime_context()` with distinct tenant IDs and then `await asyncio.sleep(0)` to force a task switch.
- `contextvars.ContextVar` is designed to be per-task in asyncio, so this should work without extra plumbing — the test just confirms the existing implementation doesn't accidentally break that guarantee.

mittalsonal:
## Summary

Adds direct tests for the `waggle.runtime_context` ContextVar lifecycle.

Closes #240

## What changed

- Covers the empty default context returned outside any runtime scope.
- Verifies context values are set inside `runtime_context()` and restored after exit.
- Verifies nested contexts inherit unspecified fields and restore the parent context.
- Verifies context reset still happens when the context body raises.
- Adds an async isolation test so concurrent tasks keep separate runtime context values across awaits.

## Validation

- `python -m pytest tests/test_runtime_context.py tests/test_logging_utils.py -q`
- `python -m ruff check tests/test_runtime_context.py`

## Summary by CodeRabbit

* **Tests**
  * Added coverage for runtime context behavior, including default/empty values, temporary overrides within a context block, nested context inheritance and field overriding, automatic reset after exceptions, and isolation across concurrent async tasks.

### DIFF
--- a/tests/test_runtime_context.py
+++ b/tests/test_runtime_context.py
@@ -48,11 +48,11 @@ def test_runtime_context_resets_when_body_raises() -> None:
 
 @pytest.mark.asyncio
 async def test_runtime_context_is_isolated_between_async_tasks() -> None:
-    async def worker(request_id: str) -> tuple[str, str]:
-        with runtime_context(request_id=request_id):
-            before_sleep = get_runtime_context().request_id
+    async def worker(tenant_id: str) -> tuple[str, str]:
+        with runtime_context(tenant_id=tenant_id):
+            before_sleep = get_runtime_context().tenant_id
             await asyncio.sleep(0)
-            after_sleep = get_runtime_context().request_id
+            after_sleep = get_runtime_context().tenant_id
             return before_sleep, after_sleep
 
     with runtime_context(request_id="parent"):
--- a/tests/test_runtime_context.py
+++ b/tests/test_runtime_context.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+from waggle.runtime_context import RuntimeContext, get_runtime_context, runtime_context
+
+
+def test_get_runtime_context_returns_empty_default_context() -> None:
+    context = get_runtime_context()
+
+    assert context == RuntimeContext()
+
+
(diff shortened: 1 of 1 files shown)


mittalsonal:
@coderabbitai review

Abhigyan-Shekhar (approved):
Rechecked the refreshed head. The runtime-context lifecycle coverage looks correct on the current diff, the earlier async-isolation concern is already resolved on this head, and the full current-head CI matrix is green. Approving.

PR #510, opened 2026-07-01
fix(recursive_context): deduplicate random ablation subqueries

### LINKED ISSUE #507: [bug]  Random ablation subqueries produce duplicates causing redundant graph retrievals
### Summary

 In src/waggle/recursive_context.py (lines 227–247), when the random_subqueries ablation flag is set, the random slicer generates word slices from the query but doesn't track which substrings have already been generated. For short queries, all max_subqueries entries are identical. Each duplicate triggers a separate graph retrieval call, wasting compute.

### Steps to reproduce

Run a test script mimicking the random slicer loop with a short query .
Print the generated subqueries
Observe that all generated subqueries are identical 

### Expected behavior

The while loop should track unique substrings using a set and only append genuinely distinct subqueries, breaking early if unique possibilities are exhausted.

### Actual behavior

 All 6 generated subqueries are identical, causing 6 identical graph retrieval calls instead of 1.

### Waggle version

0.0.1

### Install method

pip

### Environment

Windows 11 Pro, Python 3.11, VS Code with Waggle extension

### Operating System

Windows

_No response_

### Screenshot

Image

### Logs

_No response_

Prakhar54-byte:
Track seen substrings in a set so that the random slicer only appends distinct subqueries.  For short queries (e.g. 2 words) this prevents N identical graph retrieval calls when only 1 unique slice exists. The existing attempts guard still bounds total iterations.

Fixes: #507 
## Summary

- What changed?
- Why was it needed?

## Testing

- [x] `ruff check src/ tests/`
- [x] `ruff format --check src/ tests/`

### Test report

Paste the output of `pytest -v` for the files you added or changed (not the full suite, just yours):

```text
<paste here>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recursive context generation so random-substring decomposition avoids adding duplicate substring variants, reducing repeated material in generated context.
  * Resulting generated contexts are cleaner and more consistent, with fewer duplicate query substrings appearing across decomposition steps.

### DIFF
--- a/src/waggle/recursive_context.py
+++ b/src/waggle/recursive_context.py
@@ -229,6 +229,7 @@ class RecursiveContextController:
             rng = random.Random(ablation.random_seed)
             words = query.split()
             subqueries: list[RecursiveSubquery] = []
+            seen_substrings: set[str] = set()
             attempts = 0
             while len(subqueries) < max_subqueries and attempts < max_subqueries * 10:
                 attempts += 1
@@ -237,6 +238,9 @@ class RecursiveContextController:
                 slice_len = rng.randint(2, min(4, len(words)))
                 start = rng.randint(0, len(words) - slice_len)
                 substring = " ".join(words[start : start + slice_len])
+                if substring in seen_substrings:
+                    continue
+                seen_substrings.add(substring)
                 subqueries.append(
                     RecursiveSubquery(
                         query=substring,


Abhigyan-Shekhar (changes requested):
This is not ready to merge in its current form. The intended fix in `src/waggle/recursive_context.py` is small, but the PR also brings in broad unrelated churn across dozens of scripts, `templates/waggle-plus`, and a full `uv.lock` addition. That makes the review surface much larger than issue `#507` and increases regression risk for no clear benefit. Please narrow this branch to the actual recursive-context fix and any directly necessary test coverage, drop the unrelated file churn, then update/rebase so the full required CI suite runs on the cleaned-up head.

Abhigyan-Shekhar (commented):
I rechecked the current head and there still has not been a code update after the maintainer `REQUEST_CHANGES` review. This PR remains blocked for the same reason: the recursive-context fix is small, but the branch still carries large unrelated formatting churn across many scripts, `templates/waggle-plus`, and a full `uv.lock` addition. Please narrow the branch to the actual fix and any directly necessary tests, then rerun the full required CI suite on the cleaned-up head.

Abhigyan-Shekhar (approved):
The branch is now narrowed to the intended recursive-context fix only. The dedupe logic is scoped, keeps the existing attempt guard, and avoids repeated random substring subqueries without changing the rest of context assembly. Approved pending the current-head CI run finishing green.

PR #543, opened 2026-07-05
Harden Waggle Codex plugin unsigned release flow

Abhigyan-Shekhar:
## Summary

Hardens the Waggle Codex plugin release path for a repo-hosted unsigned v1 distribution, based on the current Codex marketplace behavior of installing a single marketplace/plugin bundle rather than a custom platform-index flow.

## Changes

- Keeps v1 distribution as a single multi-platform marketplace bundle and documents that per-platform artifacts should wait until Codex supports platform-specific marketplace resolution.
- Builds one-file bundled runtimes for `darwin-arm64`, `darwin-x86_64`, `linux-x86_64`, `linux-aarch64`, and `win32-x86_64`.
- Adds strict packaging validation for expected runtimes, runtime size, executable bits, bundle metadata, release metadata, checksums, and provenance attestations.
- Restores Unix execute bits after artifact download before assembling the marketplace bundle.
- Defaults manual release workflow runs to unsigned mode via `enable_signing=false`, while preserving optional macOS/Windows signing gates when credentials are later configured.
- Improves launcher diagnostics with OS, arch, runtime path, DB path, and exit details.
- Adds pre-mutation SQLite integrity validation so corrupted local memory DBs fail clearly without being modified.
- Updates Codex install/runtime docs for local-first storage, unsigned OS warnings, checksum/attestation verification, upgrade behavior, and rollback/yank guidance.

## Verification

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-codex-plugin-runtime.yml"); puts "workflow yaml ok"'`
- `.runtime-build-venv/bin/python -m pytest tests/test_package_codex_plugin.py tests/test_graph.py` -> 88 passed
- `git diff --check`
- Release workflow passed on this branch: https://github.com/Abhigyan-Shekhar/Waggle-mcp/actions/runs/28752577295
- CI logs confirm unsigned behavior: macOS and Windows signing/signature verification steps skipped because signing is disabled.
- CI assembled and uploaded final Codex plugin bundles after bundle attestation generation and verification.

## Notes

Apple Developer ID, notarization credentials, and Windows signing certificates are intentionally not required for this v1 unsigned path. If those credentials are added later, the existing workflow hooks can be enabled with `enable_signing=true`.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable signing flow for runtime releases (signing disabled by default on manual runs) with optional signature verification and build provenance attestations.
  * Release bundles now ship an additional release manifest, preserve executable permissions, and enforce stricter bundle/version consistency.
* **Bug Fixes**
  * Startup now detects corrupted local databases and refuses to proceed without modifying existing data.
  * Launcher failures now include richer platform/runtime/environment diagnostics.
* **Documentation**
  * Updated runtime build/release guidance, installation verification steps, and signing/notarization expectations.
* **Tests**
  * Expanded coverage for corrupted database handling and packaging/manifest/executable-permission validation.


Amlan-git (commented):
The release hardening looks solid overall, but there is one versioning issue to resolve before proceeding ahead. Both plugin manifests are downgraded from 0.1.0 to 0.0.1 to match pyproject.toml, while the published v0.1.17 bundle already advertises plugin version 0.1.0. This breaks monotonic plugin versioning and can prevent an existing installation from recognizing the next bundle as an upgrade.

Please establish the intended plugin version source and keep it monotonic, either by validating against the release tag or by bumping all release metadata together without downgrading the shipped plugin version.

Abhigyan-Shekhar:
Current head looks much better: the plugin manifest downgrade concern appears addressed with both plugin manifests at `0.1.1`, and CI is green. Since this PR is owner-authored, I am not merging it with only owner/self review. Requested an independent maintainer review; once that approval is present, this can come back to the merge lane.

Amlan-git (approved):
The release hardening looks solid, and I don’t see any merge-blocking concerns. The plugin versioning concern is addressed with both manifests now at 0.1.1, and the packaging/runtime validation improvements are well scoped.

Follow-up: consider adding a packaged-runtime smoke test that performs a real MCP stdio initialize, not only --server-info, so the release workflow covers the actual plugin startup path.

Approved.

PR #546, opened 2026-07-08
fix(abhi): bound archive member imports

Abhigyan-Shekhar:
## Summary

Adds bounded reads for `.abhi` ZIP archive members before decompression.

## Changes

- Checks ZIP central-directory `file_size` before reading archive members.
- Checks manifest-declared data member sizes before loading payloads.
- Bounds manifest, signature, and public-key member reads.
- Validates decrypted/plaintext member sizes and manifest size consistency.
- Adds regression tests for oversized ZIP metadata and manifest-declared oversized payloads.

## Validation

- `/Users/abhigyanshekhar/Desktop/MCP/.venv/bin/python -m pytest tests/test_graph.py -k abhi`
- `/Users/abhigyanshekhar/Desktop/MCP/.venv/bin/python -m ruff check src/waggle/abhi.py tests/test_graph.py`
- `/Users/abhigyanshekhar/Desktop/MCP/.venv/bin/python -m py_compile src/waggle/abhi.py tests/test_graph.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `.abhi` import security by enforcing size limits on manifests, signatures, keys, and archive payloads.
  * Archives with oversized, missing, or incorrectly declared members are now rejected before excessive data is read or decompressed.
  * Added validation for encrypted and unencrypted payload sizes, including invalid or mismatched declarations.
  * Signed archives missing required signature data are now rejected with a clear validation error.


Abhigyan-Shekhar:
Security fix is pushed and locally validated. GitHub rejected self-approval (`Review Can not approve your own pull request`) and admin merge is still blocked by branch protection (`New changes require approval from someone other than the last pusher` plus required checks). Remaining action: another collaborator needs to approve PR #546 after CI finishes, then it can merge.

Amlan-git (approved):
The .abhi import hardening looks good. The new bounded reads cover the main archive members, manifest, signature/public-key files, encrypted payload size, and manifest-declared size consistency without broadening the scope.

One small follow-up: for signed archives, consider failing explicitly if signatures/content.ed25519 or signatures/public_key.pem is missing, instead of letting that surface later as empty signature/key data.

Approved.

Abhigyan-Shekhar:
CI is green and the overall archive-size hardening is useful, but I cannot merge this head because the latest review surfaced a real correctness blocker.

`_read_archive_member_bounded()` currently returns `b""` for a missing member, which makes a missing file indistinguishable from a present-but-empty member. That lets `_read_member()` return before validating a nonzero manifest-declared size, and it also changes signed archive behavior: when `signatures.present` is true, missing signature/public-key members now become empty bytes instead of failing clearly.

Please use a distinct missing-member sentinel, validate missing payloads against declared sizes, and require signature/public-key members when the manifest says signatures are present. The smaller size parsing issue is also worth fixing: reject booleans and fractional floats instead of silently coercing them with `int(...)`.

Amlan-git (changes requested):
Thanks for keeping this branch updated. I rechecked the latest head, and the two previous CodeRabbit comments still look unresolved.

`_coerce_manifest_member_size()` still uses `int(value)`, so boolean and fractional manifest sizes can be silently accepted/truncated.

`_read_archive_member_bounded()` still returns `b""` for missing archive members, which keeps missing members indistinguishable from present-but-empty members. That also means nonzero manifest-declared sizes may not be validated, and signed archives can still defer missing signature/public-key failures instead of failing clearly while loading.

Could you please address those two points and add the small regression coverage CodeRabbit requested? After that, I can re-review and clear the old conversations.

Amlan-git (approved):
Thanks for the follow-up @Abhigyan-Shekhar . The previous archive-member concerns look addressed now: manifest sizes reject boolean/fractional values, missing members are distinct from empty payloads, non-empty declared missing payloads fail clearly, and signed archives now require the signature/public-key members when `signatures.present` is true.

The added regression coverage matches those cases, and CI is green. I don’t see any remaining merge-blocking concerns.

Approved.

PR #565, opened 2026-07-17
Perf: Batch embedding calls during ingestion/decomposition in graph backends

### LINKED ISSUE #124: perf(graph): batch single embed() calls during multi-node observation ingest
## Context

When Waggle processes an observation (a transcript turn), the extraction step produces multiple nodes. Each node is stored via add_node (graph.py:2418+), which calls self.embedding_model.embed(node.content) individually (graph.py:1643, called inside the observation flow). If the extraction yields 5 nodes, that is 5 serial round-trips through the model.

embed_batch() already exists (embeddings.py:175-196) and handles batched encode with proper batch_size tuning. Nobody in graph.py calls it. The batch path is 3-8x faster for multiple texts because the model processes them in a single forward pass.

## What to do

Identify the observation/ingest code path where multiple nodes are created from a single request. Pre-compute all embeddings in one embed_batch() call, then pass the pre-computed vectors to each add_node call.

The tricky part: add_node currently computes the embedding internally. You need to either:
- Add an optional `embedding` parameter to add_node so the caller can pass a pre-computed vector (check if this already exists in the signature, it likely does at graph.py:2418).
- Or extract the embedding computation out of add_node into the caller.

Check that the dedup logic (which uses the embedding for cosine similarity at graph.py:2458) still receives the correct vector.

## Acceptance criteria

- When an observation produces N nodes, embed_batch is called once with N texts instead of embed() N times.
- Dedup still works correctly (same cosine similarity results).
- A test ingests a multi-node observation and verifies all nodes got embedded.
- Add a timing assertion or log showing the batch path is faster (optional but welcome).

## Compatibility notes

This changes the observation flow in graph.py. It does NOT change the embeddings.py interface. Compatible with ONNX migration (embed_batch works with any backend). Compatible with connection pooling (separate concern). If someone is working on the async embedding sub-issue, coordinate on where the batch call happens.

## Pointers

- Source: `src/waggle/graph.py:2418` (add_node, check for existing embedding param), graph.py:1643 (where embed is called inside the node store path)
- Batch API: `src/waggle/embeddings.py:175-196` (embed_batch)
- The server-side observation entry point is in `src/waggle/server.py` (look for the observe/ingest tool handler)

## How to claim and report

1. What you changed and the code path you modified.
4. Before/after timing for a 5-node observation if you can measure it.

turfin-logic:
Resolves #124 by using embed_batch to optimize LLM embedding operations before sequential node creation.

## Summary by CodeRabbit

* **Improvements**
  * Improved the speed and efficiency of storing decomposed memory items by generating embeddings in batches.
  * Reduced delays when observing conversations by reusing precomputed embeddings for candidate items.
  * Improved topic retrieval with cached results and an option to force recomputation.
  * Expanded ABHI export controls, including scope, date filtering, redaction, signing, confidence filtering, dependency inclusion, and embedding options.
  * Preserved reliable behavior through fallback processing when batch embedding is unavailable or fails.


Abhigyan-Shekhar (approved):
Logic looks good after the latest update. The batch-embedding length guard preserves fallback behavior when the backend returns an unexpected result shape, and the precomputed embedding path keeps the existing add_node behavior for callers that do not pass an embedding. Waiting on refreshed CI before merge.

Abhigyan-Shekhar (changes requested):
CI is currently failing on the refreshed branch: `Lint (ruff)` is red on the latest head. Kindly fix the lint failure and push an update before this can be merged.

PR #502, opened 2026-06-29
feat: add community cache and staleness invalidation on graph mutation

Kirtan-pc:
## Summary

Fixes Issue #458 

### What changed?

- Added two new columns to the `tenants` table:
  - `communities_stale` (INTEGER, default `1`) — dirty flag that marks whether the cached partition is outdated
  - `cached_communities` (TEXT) — serialised JSON `{node_id: cluster_id}` partition map
- Added `_mark_communities_stale(connection)` private helper on both `MemoryGraph` (SQLite) and `Neo4jMemoryGraph` that flips `communities_stale = 1` for the current tenant
- Hooked `_mark_communities_stale()` into **every** graph-mutating code path:
  - `add_node` (fresh insert **and** dedup-merge branch)
  - `update_node`
  - `add_edge`
  - `update_edge`
  - `delete_edge`
  - `delete_node`
  - `clear_session` / `clear_project` / `clear_all` (via `_clear_scope_rows`)
  - `canonicalize_node`
  - `prune_retention_policy` (only when ≥1 node or edge is actually deleted)
- Updated `get_topics()` in both `TraversalMixin` (SQLite) and `Neo4jMemoryGraph` to:
  1. Read `communities_stale` + `cached_communities` from the tenant record
  2. Deserialise and **validate** the cached partition (key-set must match live node IDs exactly)
  3. If valid → return clusters immediately, **skipping** the heavy Louvain / greedy-modularity pass
  4. If stale or invalid → run full `_build_topic_partition`, then persist the result and clear the flag
- Added `force_recompute: bool = False` keyword argument to `get_topics()` to let callers bypass the cache unconditionally
- Exposed `force_recompute` in the `get_topics` MCP tool `inputSchema` and wired it through the server handler in `server.py`
- Added a backward-compatible `ALTER TABLE` migration in `_migrate_legacy_schema` so existing databases gain the two columns without data loss

### Why was it needed?

`get_topics`  called `community.best_partition` or `greedy_modularity_communities` on the **full tenant graph on every invocation** — an O(n·m) Louvain pass that becomes prohibitively expensive on large graphs queried frequently from the UI.

The fix introduces a **staleness signal**: whenever a node or edge is mutated the flag is set; `get_topics` checks it and only runs the heavy pass when something has actually changed. Unchanged graphs skip the Louvain pass entirely. A full recompute is still available via `force_recompute=True`, and the cached result is automatically invalidated the next time any node or edge changes.

---

## Testing

[...]

PR #575, opened 2026-07-20
perf(graph): cache get_topics results with a staleness flag

### LINKED ISSUE #451: performance(graph): cache get_communities results with a staleness flag
## Summary
`get_communities` runs a GROUP BY on every call. For large graphs and frequent UI loads this is wasteful when communities have not changed since the last `recompute_communities`.

## Task
- Cache the community summary and invalidate it when nodes/edges mutate (or when `recompute_communities` runs).
- Keep correctness: a mutation must mark the cache stale.

## Acceptance criteria
- Repeated `get_communities` calls avoid the GROUP BY when unchanged; tests cover invalidation.

payalrvs3:
Fixes #451

## Summary
- Race condition (originally flagged as a blocker here): get_topics()'s cache write was unconditional, so a mutation landing mid-recompute could get silently overwritten by a stale payload marked fresh. Added a communities_version counter on tenants, bumped by _mark_communities_stale; the cache write now only lands if the version hasn't moved since it was observed.
- Import-path invalidation (CodeRabbit's finding): import_markdown_vault, import_graph_backup, and import_abhi write nodes/edges directly with no cache hook. Each now marks the cache stale once at the end of its transaction, if it actually created or updated anything.

## Testing
```
WAGGLE_MODEL=deterministic pytest tests/test_graph.py -q
ruff check src/ tests/
ruff format --check src/ tests/
```
96 passed (4 new), ruff clean on both commands.

## Checklist
- [x] Linked issue (#451)
- [x] Focused scope - 2 source files + tests, no unrelated changes
- [x] Docs - no user-facing behavior change (force_recompute contract unchanged)
```


Abhigyan-Shekhar (changes requested):
The invalidation coverage looks much better, but there is still a correctness blocker before merge. `get_topics()` reads graph state, computes clusters, and then writes the cache as fresh without coordinating against mutations on another connection. A concurrent node or edge mutation can mark the cache stale while recomputation is in progress, after which this code can overwrite that stale marker with an old payload marked fresh. Please make recompute plus cache publication atomic, or add a mutation revision/staleness recheck immediately before publishing and discard/recompute if it changed. Keep the existing cache-hit and empty-graph behavior.

ard12 (changes requested):
Rechecked the current head (157ffec6). The atomic recompute-plus-publish
approach is the right direction and the cache invalidation coverage is
improved, but there is one correctness blocker and two follow-up items
before this is merge-ready.

**Blocker — data loss in `mutation.py` dedup projections**

The node SELECT projections at `mutation.py` lines 971–983, 985–993, and
1135–1143 omit `aliases`, `embedding_model_id`, `embedding_dim`, and
`source_turn_pair_id`. When `_row_to_node` builds a node from those rows
it substitutes empty defaults for the missing fields, and
`_merge_duplicate_node` then persists those empty defaults, silently
erasing existing aliases and turn-pair linkage on every dedup merge. This
is a data integrity regression introduced by this PR and must be fixed
before merge. Add those four columns to each of the three projections so
the fields survive a dedup round-trip.

**Follow-up 1 — docstring overstates the cache-refresh guarantee**

The comment around `traversal.py` line 1439–1441 implies recomputation
always refreshes the cache. Since `_write_topics_cache` is now skipped
when a concurrent mutation wins the version check, the docstring should
say that recomputation returns a fresh result but the cache write only
occurs when no newer mutation has invalidated it in the interim.

**Follow-up 2 — missing node-only import test**

`test_get_topics_cache_invalidated_by_markdown_vault_edge_sync` covers
edge sync but not a node-only import. Please add a test that runs a
node-only import and asserts that a previously fresh topics cache becomes
stale afterward, to ensure the invalidation path is exercised for both
node and edge mutations.

Please fix the `mutation.py` projection blocker, update the docstring,
add the node-only test case, and rerun the full required checks on the
updated head.

Abhigyan-Shekhar (changes requested):
I rechecked this after the latest main update. This PR is not merge-ready in its current form.

Main now already contains the merged community-topic caching implementation from #502, including cache invalidation and the public `get_topics(force_recompute=...)` contract. This PR is now conflicting with main and introduces a parallel topic-cache design with a different public option name, `force_refresh`, so merging it as-is would duplicate or fight the implementation already on main.

Please rebase on current main and narrow the PR to any remaining issue #451 gap that is not already covered by the merged implementation. If there is no remaining gap, this PR can be closed in favor of the merged #502 work.

payalrvs3:
Thanks for checking on this! Confirmed #502's on main get_topics(force_recompute=...) via tenants.communities_stale/cached_communities. Base caching's solid, no complaints there.

Two things from #451 are still open on main though, both already flagged in review on this PR:
- The race condition you called out as a blocker here originally - main's cache write still has no concurrency guard, so a mutation landing mid-recompute can get clobbered the same way.
- The import-path gap CodeRabbit caught - markdown vault, graph backup, and .abhi imports all write nodes/edges directly with zero cache invalidation on main.

Going to rebase onto current main and force-push a much smaller version of this branch - same PR, just a narrow diff covering those two, built on #502's design instead of the parallel one this had.

Abhigyan-Shekhar (approved):
Re-reviewed the latest cache implementation. Versioned invalidation and guarded cache writes are covered, no unresolved current-head review threads remain, and full CI passes.

PR #577, opened 2026-07-20
Fix authentication bypass for protected HTTP routes

### LINKED ISSUE #192: security(http): require API keys before resolving tenant on HTTP routes
src/waggle/server.py:2949-2995, 3075-3779

The HTTP app falls back to tenant_id from the query string whenever X-API-Key is missing, and _require_http_scope() only enforces scopes when a principal exists. That means the Starlette routes can be called without credentials and still operate on a tenant-selected graph.

Impact: a network caller can read or mutate graph state, export data, and access admin endpoints without presenting an API key. In a deployed HTTP configuration this becomes a full tenant-bound authorization bypass.

Debarati-06-ai:
## Summary

This PR fixes an authentication bypass affecting protected HTTP routes.

Previously, `_require_http_scope()` only enforced scope checks when a principal was present. If a request omitted the `X-API-Key` header, the request could continue without authentication.

## Changes

- Require an authenticated principal before enforcing scopes.
- Raise `AuthenticationError` when `X-API-Key` is missing on protected HTTP routes.

## Testing

- Ran the existing server test suite:

```text
python -m pytest tests/test_server.py

89 passed
```text
Fixes #192

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests missing an API key are now rejected with a clear 401 authentication error.
  * Protected API routes now consistently enforce authentication before processing requests.
  * Authenticated requests continue to be checked against required access permissions.
  * Authentication behavior is more consistent across graph, retention, and audit endpoints.

### DIFF
--- a/src/waggle/server/mcp.py
+++ b/src/waggle/server/mcp.py
@@ -14,12 +14,6 @@ try:
     from mcp.server.lowlevel.server import request_ctx
 except ImportError:
 
-    class _MissingRequestContext:
-        def get(self) -> Any:
-            raise LookupError
-
-    request_ctx = _MissingRequestContext()
-
 from waggle import __version__
 from waggle.config import AppConfig
 from waggle.embeddings import STATUS_READY
--- a/src/waggle/server/mcp.py
+++ b/src/waggle/server/mcp.py
@@ -13,6 +13,11 @@ from mcp.server.models import InitializationOptions
 try:
     from mcp.server.lowlevel.server import request_ctx
 except ImportError:
+    class _MissingRequestContext:
+        def get(self) -> Any:
+            raise LookupError
+
+    request_ctx = _MissingRequestContext()
 
 from waggle import __version__
 from waggle.config import AppConfig
--- a/src/waggle/server/routes.py
+++ b/src/waggle/server/routes.py
@@ -289,8 +289,9 @@ def create_http_application(app_server: WaggleServer, config: AppConfig) -> Star
         request: Request, required_scope: str, *, tenant_override: str = ""
     ) -> tuple[Any, Any | None]:
         graph, principal = _graph_from_request(request, tenant_override=tenant_override)
-        if principal is not None:
-            principal.require_scope(required_scope)
+        if principal is None:
+            raise AuthenticationError("Missing X-API-Key header.")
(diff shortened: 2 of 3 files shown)


Abhigyan-Shekhar (changes requested):
The runtime auth check is the right direction, but the latest type-hint fix was applied to the wrong function. `_graph_from_request()` can still return `(graph, None)` when no `X-API-Key` is provided, so its principal return type must stay optional. Please move the non-optional return type to `_require_http_scope()` instead, where the `principal is None` branch now raises before returning. Also add or point to a targeted server test that a protected HTTP route without `X-API-Key` is rejected.

Abhigyan-Shekhar (changes requested):
I rechecked the latest head. The type annotation issue is fixed now: `_graph_from_request()` is optional again, and `_require_http_scope()` has the non-optional principal return type.

This still is not merge-ready because the test coverage blocker remains. The PR changes an auth bypass on protected HTTP routes, but I still do not see a targeted server/route test proving that a protected HTTP request without `X-API-Key` is rejected instead of continuing. Please add a focused regression test for a protected HTTP route with no API key and rerun full CI.

Abhigyan-Shekhar (changes requested):
Thanks for addressing the authentication bypass. The auth change itself is correct, but the latest conflict resolution duplicated `_MissingRequestContext` and the `request_ctx = _MissingRequestContext()` assignment in `src/waggle/server/mcp.py`. Please remove the duplicate block and update the branch against current `main`, then rerun CI.

Abhigyan-Shekhar (changes requested):
Thanks for cleaning up the request-context fallback. The authentication change still needs one integration piece before it is safe to merge: `_require_http_scope()` now rejects every `/api/graph/*` request without `X-API-Key`, but the bundled Graph Studio API client does not send that header or provide a way to configure one. As a result, `waggle-mcp graph-studio` can load its HTML and then all graph reads/writes return 401. This also conflicts with the current local-first documentation that says local use and the local Graph Studio require no API key. Please either add a secure Graph Studio credential/session flow and update the docs, or preserve the intentional keyless loopback-local path while requiring keys for remotely reachable deployments. A browser-level regression test should cover the chosen behavior.

PR #586, opened 2026-07-31
fix: Add memory usage statistics to the waggle-mcp doctor command

### LINKED ISSUE #557: Add memory usage statistics to the waggle-mcp doctor command
### Problem or use case

The `waggle-mcp doctor` command verifies the installation and configuration of Waggle, but it does not provide any information about the current memory database.

Users cannot quickly check how much memory has been stored, how many nodes or edges exist, or which backend and embedding model are currently being used. This makes it harder to verify that Waggle is actively storing and retrieving memory.

### Proposed solution

Extend the `waggle-mcp doctor` command to include a "Memory Statistics" section.

Suggested information:
- Total graph nodes
- Total graph edges
- Number of stored conversations
- Database size
- Active backend (SQLite or Neo4j)
- Active embedding model

Example:

Memory Statistics
-----------------
Nodes: 1284
Edges: 2947
Conversations: 156
Database Size: 18.2 MB
Backend: SQLite
Embedding Model: all-MiniLM-L6-v2

This would provide users with a quick health check and make debugging easier.

### Alternatives considered

Users can inspect the database manually using SQLite tools or custom scripts, but this requires technical knowledge and is inconvenient for everyday use.

Providing these statistics directly through `waggle-mcp doctor` offers a faster and more user-friendly experience.

### Suggested implementation scope

Possible files:
- src/waggle/cli/doctor.py
- src/waggle/storage/
- src/waggle/server.py
- tests/test_doctor.py

The implementation would gather statistics from the configured backend and display them as part of the existing doctor output.

### Screenshot or mockup

$ waggle-mcp doctor

✓ MCP configuration found
✓ SQLite database connected

Memory Statistics
-----------------
Nodes: 1284
Edges: 2947
Conversations: 156
Database Size: 18.2 MB
Backend: SQLite
Embedding Model: all-MiniLM-L6-v2

Status: Healthy ✅

webbrain-one:
Closes #557

## Summary by CodeRabbit

* **New Features**
  * Added memory statistics to the `doctor` diagnostics, including database size, graph counts, transcript records, sessions, context windows, backend, and embedding model.
  * Non-SQLite backends now clearly indicate when file statistics are unavailable.

* **Bug Fixes**
  * Improved diagnostics handling for missing database tables and database errors.

* **Tests**
  * Added coverage for memory statistics in both text and JSON diagnostic output.


Abhigyan-Shekhar (changes requested):
I reviewed the current head and this is not merge-ready yet.

The PR adds a new standalone `src/waggle/cli/doctor.py`, but the actual doctor command used by this repo is still wired through `waggle.server._run_doctor(...)` and covered by the existing `tests/test_server.py` doctor tests. I do not see this new `print_memory_stats(...)` function imported or called by the real doctor command path, so merging this would add dead code without changing `waggle-mcp doctor` output.

There are also two contract issues in the new helper itself: it hardcodes the fallback database path and backend instead of using the resolved `AppConfig` values, and it queries a `conversations` table that does not match the current SQLite schema used by Waggle transcripts.

Please wire the memory statistics into the existing doctor implementation, use the resolved config/backend as the source of truth, query the current schema tables, and add/adjust doctor tests proving the output is present in the real CLI path.

Amlan-git (approved):
Thanks for the follow-up. The latest commit resolves the earlier wiring, configuration, and schema concerns. The memory statistics now run through the real doctor path, the SQLite behavior is covered, and CI is green. I don’t see any remaining merge-blocking concerns.
One non-blocking follow-up: for non-SQLite backends, it would be clearer to display the statistics as unavailable rather than showing zero-valued counters with an ok status. The conversation count may also eventually want to use the full tenant/project/agent/session scope rather than session_id alone.
Approved.

Abhigyan-Shekhar:
Follow-up review is clean from my side. Current head has all required doctor-memory-stats fixes wired through the real doctor path, CI is green, CodeRabbit is addressed, and the remaining CHANGES_REQUESTED state appears stale from the pre-fix review. This is merge-ready.

Amlan-git (dismissed):
Re-reviewed the latest head (`d9877db`) against the current `main`, including the commit history, outstanding review threads, changed files, and CI results.

The implementation now addresses the substantive concerns raised against the initial revision:

- Memory statistics are wired into the actual `waggle-mcp doctor` execution path rather than remaining as unused collection logic.
- Statistics are derived from the active `AppConfig`, so the configured backend, embedding model, and database path are used consistently.
- The new `memory_stats` result is included in both human-readable and JSON output without disrupting the existing diagnostic summary or exit-code behavior.
- The SQLite implementation reports nodes, edges, transcript records, conversation sessions, context windows, and database size.
- Missing tables degrade to zero counts, while SQLite connection/query failures produce a warning result instead of incorrectly failing the entire doctor command.
- The tests use a real temporary `MemoryGraph` and verify that stored data is reflected in the doctor output, alongside the JSON schema integration.

I also compared this head with the previously reviewed `144bb37` revision. The latest commit is a merge from `main`; it does not modify the PR-owned implementation in `src/waggle/server/cli.py` or `tests/test_server.py`. Therefore, the merge does not invalidate the earlier functional review or reintroduce the original wiring/configuration problems.

Verification performed:

- All 17 current GitHub checks are successful, including the supported Linux and Windows Python matrices, lint, packaging, repository validation, and security-oriented checks.
- All 9 doctor-focused tests pass locally against the exact PR head.
- Ruff lint and formatting checks pass for both changed files.
- A broader `tests/test_server.py` run produced 90 passing tests and two unrelated graph-editor failures caused by the locally installed MCP SDK rejecting the repository’s existing `Server(..., on_list_tools=...)` API. Those failures are outside this diff, and the repository’s supported CI environments are green.

I do not see a merge-blocking correctness or regression issue.

There are a few improvements worth considering, but I would treat them as follow-up work:

1. **Non-SQLite reporting:** For a Neo4j configuration, `_collect_memory_stats()` currently returns `status: "ok"` with zero-valued SQLite statistics. The explanatory `reason` is retained in JSON but is not shown in normal text output. That can make an unsupported result look like an empty memory store. A future change should use an explicit `unavailable` or `warn` state and avoid displaying SQLite-specific counts when the active backend is not SQLite.

2. **Filesystem error handling:** The initial `Path.exists()` and `Path.stat()` calls occur before the guarded SQLite block. A permission error or path race could therefore terminate `doctor` instead of being reported as a diagnostic warning. Wrapping the filesystem inspection and database access in the same error boundary would make this command more resilient.

3. **Database-size presentation:** Small SQLite files can display as `0.0 MB`. Falling back to KiB below 1 MiB would make the reported value more useful, especially for newly created databases.

4. **Additional regression coverage:** The current tests cover the principal SQLite path and JSON integration. Coverage could be strengthened with explicit assertions for edge and context-window counts, database size, unsupported backends, missing tables, and SQLite read failures.

None of these observations undermine the correctness of the primary SQLite implementation or justify holding the PR. The remaining `CHANGES_REQUESTED` state appears to come from the superseded initial revision; the underlying blocking concerns have been addressed.

**Approved. This is merge-ready from my side, with the points above suitable for non-blocking follow-up work.**

Amlan-git (approved):
Re-reviewed the latest head. The doctor memory-statistics implementation is correctly wired through the real CLI path, and the new hardening addresses the remaining concerns: unsupported backends are reported as unavailable, filesystem and SQLite failures degrade safely, SQLite is opened read-only, and small database sizes are now useful rather than rounding to 0.0 MB.
I don’t see any merge-blocking code concerns. The latest CI and HOL Scanner workflows are currently waiting for approval and have not executed, so they should be run and confirmed green before merge.
One optional cleanup: for non-SQLite backends, avoid printing the local db_path size before reporting that memory statistics are unavailable. That does not need to hold this PR.
Approved.

Abhigyan-Shekhar (approved):
Re-reviewed the hardened doctor statistics implementation. Unsupported backends and SQLite read failures now report unavailable state correctly, filesystem/database access is guarded, and full current-head CI passes.

PR #590, opened 2026-07-31
fix: clarify unavailable doctor memory stats

Abhigyan-Shekhar:
Follow-up to #586. This is stacked on top of #586 because the doctor memory stats code is not on main yet; once #586 merges, this PR should show only the follow-up commit.\n\nChanges:\n- Report non-SQLite doctor memory stats as unavailable with a warning/reason instead of zero-valued ok counters.\n- Count conversation sessions by distinct tenant/project/agent/session scope instead of session_id alone.\n- Add targeted tests for both behaviors.\n\nValidation:\n- uv run --extra dev python -m ruff check src/waggle/server/cli.py tests/test_server.py\n- uv run --extra dev python -m pytest tests/test_server.py -q -k "doctor"

## Summary by CodeRabbit

- **New Features**
  - Added memory statistics to the doctor diagnostic command, including backend, model, database size, nodes, edges, transcripts, sessions, and context windows.
  - Added clear warnings when memory statistics are unavailable for non-SQLite backends.
  - Included memory statistics in JSON diagnostic output.

- **Bug Fixes**
  - Improved diagnostic reporting for SQLite-backed memory configurations.


Amlan-git (changes requested):
Thanks for separating the follow-up behavior from #586. Counting conversations by the full tenant/project/agent/session scope is the right correction, and the accompanying test demonstrates why session_id alone is insufficient. Reporting non-SQLite statistics as unavailable is also clearer than presenting zero-valued SQLite counters.
I’m requesting changes for one correctness issue before this is merged.
The current head (686ebf8) is still based on an older #586 revision and does not contain the hardening now present at the latest #586 head (35cda82). In particular, the sqlite3.Error handler sets status to warn but leaves available=True. The doctor renderer then prints zero or partially collected counters as valid statistics.
I confirmed this against the exact PR head by using a SQLite path that cannot be opened as a database. _collect_memory_stats() returned a warning and an error reason, but still reported the metrics as available with zero values. This validates the unresolved CodeRabbit comment on the same path; it is a real diagnostic correctness issue rather than a theoretical edge case.
Please restack this PR onto the latest #586 implementation and keep the follow-up delta focused on:
- counting conversations by distinct tenant/project/agent/session scope;
- retaining the corresponding regression test;
- preserving #586’s guarded filesystem handling, read-only SQLite access, useful size formatting, and safe unavailable-state rendering;
- adding a direct SQLite read-failure test that verifies unavailable or partial statistics are never exposed as valid.

The existing validation is otherwise healthy: all GitHub checks on the current head pass, the full tests/test_server.py module passes locally with 90 tests, and Ruff lint and formatting checks are clean. CI should be rerun after the branch is refreshed because the current runs predate the latest base.
Once the branch is restacked and the failure path is covered, this should be a straightforward follow-up to approve.

PR #601, opened 2026-08-02
Add opt-in telemetry and project memory CLI

Abhigyan-Shekhar:
## Summary
- add a disabled-by-default anonymous telemetry system with local opt-in/out CLI controls, strict event/property allowlists, local queueing, and MCP/CLI capture boundaries
- add project-memory terminal workflow commands: `bootstrap`, `search`, `stats`, `timeline`, and `inspect-node`
- improve recursive context evidence prioritization with focused tests for count/update/detail recall cases
- document telemetry privacy guarantees and the project-memory CLI workflow

## Verification
- `PYTHONPATH=src .venv/bin/python -m pytest` in the original working tree: `1027 passed, 7 skipped`
- clean PR branch: `ruff check` on touched files passed
- clean PR branch: full suite reached `977 passed, 7 skipped`; `tests/test_stdio_integration.py::test_server_stdio_initialize_and_basic_calls` timed out in full-suite order but passed immediately when rerun in isolation (`1 passed in 1.46s`)

## Scope Notes
- benchmark outputs and LongMemEval artifacts are intentionally excluded from this PR
- `PRIVACY.md` existed on the feature branch but is deleted on current `main`; telemetry privacy details are in `docs/telemetry.md` and linked from `README.md`

## Summary by CodeRabbit

* **New Features**
  * Added terminal commands to bootstrap repositories and inspect project memory through search, statistics, timelines, and node details.
  * Added optional anonymous telemetry controls, status reporting, previews, and configuration commands.
* **Documentation**
  * Expanded CLI reference and added guides for project-memory workflows and telemetry behavior.
  * Documented supported options, output formats, privacy safeguards, and scripting examples.
* **Tests**
  * Added coverage for repository bootstrapping, CLI workflows, telemetry configuration, sanitization, retention, and failure handling.


Amlan-git (changes requested):
Thanks for tightening the scope here—the final branch is much easier to evaluate now that the unrelated recursive-context work has been removed. The telemetry privacy boundary, allowlisted payload construction, opt-in defaults, queue retention, and terminal project-memory workflow are all thoughtfully implemented, and CI is green.
I’d like two issues addressed before merge:
The configured telemetry endpoint, https://analytics.waggle.dev/v1/events, currently fails during TLS negotiation. Because delivery errors are intentionally swallowed, a user can explicitly enable telemetry and receive no indication that every event remains queued and is never delivered. Please ensure the production endpoint is deployed and TLS-valid before this ships. I’d also recommend adding a lightweight endpoint smoke check or a configurable endpoint seam so delivery can be verified without relying exclusively on mocked unit tests.

bootstrap_repository() catches individual add_node() failures, which is sensible for partial progress, but it does not report a failure count. If every candidate write fails, _run_bootstrap() still exits with status 0 after creating no memory. That gives both users and automation a false success result. Please include failed-write information in the result and return a non-zero status when candidates were found but none could be persisted, with a regression test for that case.

Non-blocking follow-ups: validate that --max-files and --max-file-bytes are positive, consolidate the duplicated telemetry command handling, and consider cross-process coordination for the shared telemetry queue.
Once the endpoint and bootstrap failure semantics are addressed, this should be in good shape to approve.

Amlan-git (commented):
Thanks for the update. The bootstrap failure handling is now in good shape: failed candidates are counted and reported, the all-write-failure path returns a non-zero status, and the regression coverage exercises both the result object and CLI behavior. Consolidating telemetry command handling also resolves the earlier duplication concern.

One blocker remains before approval. The new telemetry smoke command and WAGGLE_TELEMETRY_ENDPOINT override make endpoint failures observable and easier to diagnose, but the default endpoint shipped by the client—https://analytics.waggle.dev/v1/events—still fails TLS negotiation. As a result, normal opted-in telemetry remains unable to deliver events; they are simply retained in the local queue while background flush failures are suppressed.
Please either deploy/fix the default endpoint or change the default to a working production endpoint, then confirm that:

_'waggle-mcp telemetry smoke'_ returns a successful result without requiring an override. Once that succeeds, I’m comfortable approving the implementation.

Non-blocking follow-ups: validate positive values for the bootstrap limits, consider cross-process coordination for the shared queue, and update the PR description to remove the recursive-context work that is no longer in the final diff.

PR #605, opened 2026-08-03
fix: namespace window ids when importing ABHI snapshots

vipul674:
## Summary

- abhi_to_snapshot copied context_window_edges verbatim, so imported edges kept stale ids that did not exist in the new graph. Apply the same id namespacing used for nodes.

## Testing

- [x] `WAGGLE_MODEL=deterministic pytest -q` (targeted suites)

### Test report

```text
47 passed in 2.53s (test_ingest_transcript_handoff)
```

### DIFF
--- a/src/waggle/abhi.py
+++ b/src/waggle/abhi.py
@@ -1994,6 +1994,16 @@ def abhi_to_snapshot(
                 "tenant_id": str(manifest.get("tenant", "")) or fallback_tenant_id,
             }
         )
+    context_window_edges = []
+    for raw_edge in manifest.get("context_window_edges", []):
+        context_window_edges.append(
+            {
+                **deepcopy(raw_edge),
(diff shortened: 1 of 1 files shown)


vipul674:
👋 @Abhigyan-Shekhar — PR is ready for review. Please take a look when you get a chance. Thanks!

Abhigyan-Shekhar (changes requested):
Thanks for fixing the stale context-window edge identifiers. The implementation is focused, but the PR currently has no regression test for this path. Please add a test that imports an ABHI document with a namespace and asserts the edge `id`, `source_window_id`, and `target_window_id` are all namespaced and still reference the imported windows. It would also be useful to cover an empty or missing edge ID so malformed input does not silently produce a dangling `namespace:` identifier.

PR #606, opened 2026-08-03
fix: make transcript turn-index allocation atomic

vipul674:
## Summary

- Two concurrent observe_conversation calls could both read the same max turn_index and assign duplicate indexes. Replace the MAX-based allocation with an atomic per-session counter node (unique constraint on tenant+session) seeded from existing transcripts.

## Testing

- [x] `WAGGLE_MODEL=deterministic pytest -q` (targeted suites)

### Test report

```text
47 passed in 2.53s (test_ingest_transcript_handoff)
```

### DIFF
--- a/src/waggle/neo4j_graph.py
+++ b/src/waggle/neo4j_graph.py
@@ -310,6 +310,10 @@ class Neo4jMemoryGraph:
                 FOR (a:GraphAuditEvent) REQUIRE a.event_id IS UNIQUE
                 """).consume()
             session.run("""
+                CREATE CONSTRAINT waggle_transcript_counter_session IF NOT EXISTS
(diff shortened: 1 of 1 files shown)


vipul674:
👋 @Abhigyan-Shekhar — PR is ready for review. Please take a look when you get a chance. Thanks!

Abhigyan-Shekhar (changes requested):
Thanks for replacing the MAX-based allocation with an atomic counter. Because concurrency is the purpose of this change, the current unrelated targeted test run is not enough to prove the fix. Please add Neo4j-stub coverage showing two allocations for the same tenant/session produce distinct monotonic turn indices, plus a migration case where existing transcript rows seed the first counter value correctly. The test should also verify the new composite constraint is created.

PR #612, opened 2026-08-09
Fix release binary doctor smoke test

Abhigyan-Shekhar:
## What changed

Updates the legacy binary release smoke check to run `doctor --help` instead of `doctor`.

## Why

The binary itself built successfully for v0.1.20, but `doctor` correctly reports a nonzero status on a fresh CI runner with no MCP config or database directory. The release smoke test should verify that the command is packaged and callable, not require a configured user environment.

## Validation

- Focused packaging tests: `21 passed`
- Ruff and diff checks passed
- The v0.1.20 Codex runtime-and-bundle workflow completed successfully for all five platforms.


Amlan-git (approved):
LGTM. 
approving

Issue #671, opened 2026-08-12
docs(errors): add troubleshooting links for common errors


Abhigyan-Shekhar:
## Goal\nAdd a compact mapping from a few common validation or connection errors to the relevant setup or security documentation.\n\n## Suggested files\n- docs/install/troubleshooting.md\n\n## Acceptance criteria\n- The change stays focused on the stated scope.\n- Documentation or tests are updated as appropriate.\n- The relevant focused test or documentation validation passes.\n\nThis is intentionally scoped as a starter contribution.

abhiramvsmg:
hai ! can i get this work to contribute ?

NaisargPurohit:
Hi Maintainer I am a student at BMSCE participating in the COMMIT-ed Open Source Sprint. I would love to work on this issue and contribute to the project! I have experience with this tech stack and have already set up the project locally.
Could you please assign this to me? Thanks!

itz1cr:
@Abhigyan-Shekhar Hi! I'm a Computer Science student looking to dive deeper into open-source, and I'd love to work on this. Could you assign this issue to me?

PR #674, opened 2026-08-12
Enhance troubleshooting documentation for API key errors

itsnevu:
This adds two small pieces to docs/install/troubleshooting.md. First, a new entry for the AuthenticationError: Invalid API key message that HTTP transport returns when a request is missing a valid X-API-Key header. It shows the waggle-mcp create-api-key command needed to issue a key and links to docs/reference.md and docs/security/security-model.md so readers can see the full admin command list and how API keys and tenant scopes work. This case was already mentioned in the top-level README's failure modes list but was missing from the dedicated troubleshooting page, so it was easy to miss.

Second, a compact table at the end of the file mapping common categories of errors, such as install and PATH issues, config or connection errors, authentication errors, and production hardening, to the most relevant setup or security doc (README install guide, environment-variables.md, security-model.md, reference.md, and hardening-checklist.md). This is the mapping requested in the issue.

This is a docs-only change. I did not run the test suite or linters since I only edited a markdown file and made no code changes; I did check that every link and command referenced already exists in the current docs and CLI. No secrets or sensitive data are included in the diff.

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting guidance for HTTP authentication failures on the `/mcp` endpoint.
  * Documented API-key header usage, key creation steps, and distinct missing, invalid, and expired key errors.
  * Added links to security, administration, configuration, and hardening guidance.
  * Clarified that health and metrics endpoints remain accessible without authentication.
  * Included a reference table connecting common errors with relevant documentation.


itsnevu:
Pushed a fix for the review comment. Two changes to the new section:

Scoped the header requirement to /mcp. The health and metrics endpoints don't require it, per tests/test_platform.py.

Split the failure into the three messages the auth path actually raises, since they need different fixes: "Missing X-API-Key header." from src/waggle/protocol/mcp/http.py, and "Invalid API key." / "API key expired." from src/waggle/auth.py.

I also dropped the "Related setup and security documentation" table from the bottom of the file. @abhiramvsmg is already adding an error-to-docs mapping table in #675 for #671, and two mapping tables in one page would just be confusing. This PR now only adds the AuthenticationError section, which #675 doesn't cover. Sorry for the overlap, I should have checked the open PRs before writing that part.

Unrelated, but worth a separate issue: README.md documents key creation as `waggle-mcp keys create`, but the actual subcommand in src/waggle/server/cli.py is `create-api-key`. Happy to open a follow-up for that if it's useful.

itsnevu:
@Abhigyan-Shekhar gentle ping on this one — it's been sitting about a week. I've just merged main in, so the branch is current and the two checks that can run are green; the remaining six are blocked on a maintainer approving the workflow run.

No rush from my side, but flagging one thing that decays with time: #675 is still open and adds its own error-to-docs table, so whichever of the two lands second will need rewriting rather than just rebasing. Happy to defer to #675 and close this if you'd prefer that shape — I'd rather not have you review two overlapping PRs.

Also still on offer from my earlier comment: the README documents `waggle-mcp keys create` but the actual command is `create-api-key`. That's a separate one-line fix, so I left it out of this PR. Say the word and I'll open it as a follow-up.

Abhigyan-Shekhar (approved):
Maintainer review complete. The HTTP authentication troubleshooting guidance matches the server behavior, all review threads are resolved, and the refreshed full CI suite is green.

PR #694, opened 2026-08-18
docs: add instructions for running individual pytest files and cases

soha-k10:
## Description
Added  concise examples for executing a single test file and an individual test case using `pytest` to `CONTRIBUTING.md`. This helps new contributors quickly validate isolated changes without running the full test suite.

## Scope & Changes
- Updated `CONTRIBUTING.md` under the testing section with focused `pytest` command snippets.
- Refenced valid repository test paths (`tests/test_...py`).

## Testing
- Verified Markdown formatting renders cleanly.
- Confirmed `pytest` commands run as expected against existing project test files.

## Summary by CodeRabbit

* **Documentation**
  * Updated testing guidance with instructions for running a single test file or a specific test case.
  * Simplified the formatting of the general test-running example.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -225,7 +225,7 @@ If your issue does not clearly require one of these files, look for a narrower p
 
 ## Running Tests
 
-```bash
+```
 # Fast: deterministic embeddings — no 420 MB download, no network
 WAGGLE_MODEL=deterministic pytest -q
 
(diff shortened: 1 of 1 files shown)


Abhigyan-Shekhar (changes requested):
The edit leaves the existing test-command fence malformed: its opening ```bash was changed to ```, then a new ```markdown block begins before the original commands are closed. As a result, the existing test commands and the new “Running Focused Tests” section render as code instead of normal documentation, and the nested bash fence is invalid. Restore the outer `bash` fence around the existing commands, close it before the prose heading, and use a regular Markdown heading with a separate `bash` fence for the examples.

PR #701, opened 2026-08-23
fix(mcp): reject unsupported POST content types

### LINKED ISSUE #640: test(http): reject unsupported request content types
## Goal\nAdd focused coverage for an unsupported Content-Type and assert the endpoint returns a clear client error without processing the body.\n\n## Suggested files\n- tests/test_mcp_http.py, src/waggle/protocol/mcp/http.py\n\n## Acceptance criteria\n- The change stays focused on the stated scope.\n- Documentation or tests are updated as appropriate.\n- The relevant focused test or documentation validation passes.\n\nThis is intentionally scoped as a starter contribution.

adity982:
## Summary
- reject MCP POST requests whose media type is not application/json
- return a structured 415 response before consuming or forwarding the body
- cover parameters such as text/plain; charset=utf-8

## Validation
- uv run pytest tests/test_mcp_http.py -q (4 passed)
- uv run ruff check src/waggle/protocol/mcp/http.py tests/test_mcp_http.py

Closes #640

## Summary by CodeRabbit

* **Bug Fixes**
  * MCP HTTP endpoints now require `Content-Type: application/json` for POST requests.
  * Requests with missing or unsupported content types return HTTP 415 with a clear error before further processing.


Amlan-git (changes requested):
Thanks for the focused update. The implementation is correctly placed before `Request.body()` and downstream MCP handling, and the structured 415 response is appropriate for unsupported or missing media types.

Before merge, please address two small items :

1. The new test confirms that the downstream MCP app is not called, but it does not prove that the request body was never consumed. As noted in the unresolved CodeRabbit review thread, `app_called is False` would still pass if `receive()` were called and the body were read before returning 415. Since issue #640 explicitly requires rejection without processing the body, please add a receive-call counter or make the test’s `receive()` function fail if invoked, then assert zero calls alongside the existing status and response-body assertions.

2. The Ruff formatting check is currently failing on the new test. Please run the formatter and push the resulting change so CI is fully green.

The runtime behavior itself looks correct and the scope remains appropriately focused. Once the body-consumption assertion is added and formatting CI passes, this should be good to approve.

adity982:
Addressed both requested items in the latest commit: the unsupported-media-type test now injects a failing receive callback and asserts it is never called, and the file passes Ruff formatting. Focused validation: uvx ruff check tests/test_mcp_http.py and uv run pytest tests/test_mcp_http.py -q (4 passed).

Amlan-git (approved):
Thanks for the follow-up. The updated test now directly proves that unsupported media types are rejected before the request body is consumed: receive() fails if invoked, zero calls are asserted, and the test still verifies the 415 response, structured error, and skipped MCP invocation.

The earlier CodeRabbit thread is resolved, its latest review found no additional actionable issues, and all checks—including Ruff formatting—are green. The change remains focused and satisfies the linked issue’s acceptance criteria.

I don’t see any remaining merge-blocking concerns.
Approved.

PR #587, opened 2026-07-31
Migrate MCP transport to SDK v2

Abhigyan-Shekhar:
## Summary

Migrates Waggle's MCP transport layer to the MCP SDK v2 low-level server while keeping the historical `WaggleServer` import surface available as a compatibility shell.

The PR separates protocol concerns from tool behavior:

- adds `waggle.protocol.mcp` for SDK v2 stdio, streamable HTTP, adapter, and shared resource/prompt surface
- extracts protocol-independent tool definitions, request context, results, validation, and dispatcher into `waggle.tools`
- keeps legacy camelCase result/schema compatibility for older local callers
- updates stdio and HTTP routes to use the SDK v2 transport path
- refreshes MCP fixture generation and compatibility surface tests
- hardens stdio smoke, distribution validation, and package validation paths

## Impact

- MCP clients use SDK v2 field names on the new transport path (`input_schema`, `structured_content`, `is_error`).
- Existing internal tests and older callers that import `WaggleServer` still get camelCase compatibility fields.
- Tool argument schemas are now validated centrally with JSON Schema before dispatch.
- HTTP `/mcp` now routes through the v2 streamable HTTP app behind the existing Waggle auth/rate-limit wrapper.

## Validation

Clean branch on top of `origin/main`:

- `uv run --with pytest-asyncio pytest -q` -> `885 passed, 7 skipped`
- `uv run --with pytest-asyncio pytest tests/test_mcp_tool_surface.py tests/test_mcp_v2_adapter.py tests/test_stdio_integration.py tests/test_packaging_metadata.py tests/test_repository_layout.py -q` -> `29 passed`
- `uv run --with ruff ruff check src/waggle/protocol/mcp src/waggle/server/mcp.py src/waggle/server/cli.py src/waggle/server/routes.py src/waggle/tools scripts/smoke_test_mcp.py scripts/check_distribution.py scripts/generate_mcp_fixtures.py tests/test_mcp_tool_surface.py tests/test_mcp_v2_adapter.py tests/test_stdio_integration.py` -> passed
- `uv run python scripts/check_distribution.py` -> passed
- `uv run python scripts/smoke_test_mcp.py` -> passed
- wheel/sdist build and installed `waggle-mcp` stdio smoke were validated in throwaway temp dirs before opening the PR

## Notes

This branch was rebuilt cleanly from `origin/main` to avoid unrelated LongMemEval/release-history commits from the original working branch.

## Summary by CodeRabbit

* **New Features**
[...]

PR #707, opened 2026-08-30
test: guard request_ctx import fallback against regression (fixes #704)

### LINKED ISSUE #704: [bug] ImportError: cannot import name 'request_ctx' from mcp.server.lowlevel.server on fresh install (waggle-mcp 0.0.1, mcp 2.1.1)
### Summary

Running `waggle --help` immediately after a fresh pip install crashes with an ImportError. waggle/server.py imports `request_ctx` from 
`mcp.server.lowlevel.server`, but this name does not exist in mcp 2.1.1, the version currently resolved by pip when installing waggle-mcp. The CLI is completely unusable out of the box for new users.

### Steps to reproduce

1. pip install waggle-mcp
2. waggle --help
3. Observe ImportError instead of the expected help output

### Expected behavior

`waggle --help` should print the list of available CLI commands and options.

### Actual behavior

Command crashes immediately with:

ImportError: cannot import name 'request_ctx' from 'mcp.server.lowlevel.server' 
(C:\Users\USER\AppData\Local\Programs\Python\Python313\Lib\site-packages\mcp\server\lowlevel\server.py)

Full traceback:
File "<frozen runpy>", line 198, in _run_module_as_main
File "<frozen runpy>", line 88, in _run_code
File "C:\Users\USER\AppData\Local\Programs\Python\Python313\Scripts\waggle.exe\__main__.py", line 4, in <module>
    from waggle.server import main
File "C:\Users\USER\AppData\Local\Programs\Python\Python313\Lib\site-packages\waggle\server.py", line 28, in <module>
    from mcp.server.lowlevel.server import request_ctx
ImportError: cannot import name 'request_ctx' from 'mcp.server.lowlevel.server'

### Waggle version

0.0.1

### Install method

pip

### Environment

Windows 11, Python 3.13, mcp 2.1.1 (installed automatically as a dependency), VS Code integrated terminal, fresh install with no prior waggle-mcp version.

### Operating System

Windows

PowerShell 5.1.26100.8875 (Windows PowerShell), execution policy: default, terminal: VS Code integrated PowerShell terminal

### Screenshot

Image

### Logs

pip show waggle-mcp:
Name: waggle-mcp
Version: 0.0.1
Summary: MCP server that gives LLMs persistent graph-structured memory
Requires: anthropic, anyio, cryptography, google-api-python-client, google-auth, 
google-genai, mcp, networkx, numpy, openai, portkey-ai, pydantic, python-dotenv, 
python-louvain, pyvis, requests, rich, sentence-transformers, starlette, uvicorn

pip show mcp:
Name: mcp
Version: 2.1.1
Summary: Model Context Protocol SDK
Requires: anyio, httpx2, jsonschema, mcp-types, opentelemetry-api, pydantic, pyjwt, 
python-multipart, pywin32, sse-starlette, starlette, typing-extensions, 
typing-inspection, uvicorn

#REFERENCE:
Image

madhav1811:
## Summary

Closes #704.

The crash reported in #704 (`ImportError: cannot import name 'request_ctx' from 'mcp.server.lowlevel.server'` on a fresh `pip install waggle-mcp` + `waggle --help`) is caused by `mcp>=2.1` dropping `request_ctx` from `mcp.server.lowlevel.server`.

I confirmed against `mcp==2.1.1` (the exact version from the report) that:

- `src/waggle/server/mcp.py` **already guards this import** with a `try/except ImportError` fallback (`_MissingRequestContext`), added during the SDK v2 migration (#587).
- `waggle --help` runs successfully on current `main` with `mcp==2.1.1` installed.
- The only version published on PyPI is `0.0.1`, which predates that fallback — so `pip install waggle-mcp` today still reproduces the crash from the issue, even though `main` already fixes it. **A new release needs to be cut** for end users to get this fix.

Since the fix already exists but was never regression-tested, this PR adds test coverage so a future `mcp` release touching the same import gets caught in CI instead of silently reproducing this crash again.

- `test_waggle_server_mcp_survives_missing_request_ctx`: removes `request_ctx` from `mcp.server.lowlevel.server`, reloads `waggle.server.mcp`, and asserts the fallback's `.get()` raises `LookupError` (not `ImportError`).
- `test_waggle_help_runs_without_request_ctx`: runs `python -m waggle.server --help` as a subprocess and asserts it exits 0.

I verified both tests fail with the exact traceback from the issue when the `try/except` fallback is temporarily removed, confirming they'd catch this regression.

## Test plan

- [x] `pytest tests/test_request_ctx_compat.py -v` passes
- [x] `pytest tests/test_server.py tests/test_mcp_http.py tests/test_mcp_v2_adapter.py tests/test_mcp_tool_surface.py -q` passes (109 passed, no regressions)
- [x] Manually reverted the `try/except` fallback locally and confirmed both new tests fail with the same `ImportError` reported in #704, then restored it and confirmed they pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the server module imports successfully when the optional request context is unavailable.
  * Prevented fallback context state from affecting subsequent tests.

* **Tests**
  * Strengthened regression coverage for request-context fallback behavior and restoration.
  * Verified compatibility with both available and unavailable request-context implementations.

* **Documentation**
  * Added a changelog entry documenting the compatibility fix and regression coverage.


Abhigyan-Shekhar (changes requested):
Thanks for adding coverage for the compatibility fallback. One test-isolation issue still needs adjustment: `test_waggle_server_mcp_survives_missing_request_ctx` reloads `waggle.server.mcp` in the `finally` block while the monkeypatch is still active, so the module remains loaded with the fallback object even after pytest restores the SDK attribute. That can leak fallback state into later tests. Please restore the original `request_ctx` first (or call `monkeypatch.undo()` in cleanup) and then reload `waggle.server.mcp`, with an assertion that the real object is restored.

madhav1811:
Pushed a fix for the test-isolation issue:

- `monkeypatch.undo()` now runs in the `finally` block **before** reloading `waggle.server.mcp`, so the reload picks up the SDK's real restored state instead of the fallback that was still active.
- Added an assertion after the reload confirming the restoration: it checks identity against the SDK's real `request_ctx` when present, or that a fresh `_MissingRequestContext` was rebuilt (not a leaked stale one) when the installed SDK genuinely lacks it.
- Also merged in latest `main` and fixed a `ruff format` flag on the added assertion.

Commits: d68037a, b692d48, 8ac40a7 (fix/gh-704-request-ctx-import-regression-test).

CI is green on all Test/Lint/smoke-test jobs. Would appreciate a re-review when you get a chance — thanks!

PR #713, opened 2026-09-01
fix(retrieval): bind CandidateT to a Protocol requiring .ts in rank_t…

### LINKED ISSUE #712: [bug] rank_temporal() accepts candidates without a `.ts` attribute — unbound generic causes runtime AttributeError instead of a type error
### Summary

`rank_temporal()` in `src/waggle/retrieval/rankers/temporal.py` declares its
candidate type as an unbound `TypeVar` (`CandidateT = TypeVar("CandidateT")`),
but the function body accesses `candidate.ts` unconditionally on line 67.
Nothing in the type signature requires candidates to have a `.ts` attribute,
so a caller can pass a sequence of objects lacking that field and get no
static warning — only a runtime `AttributeError` when the function actually
runs. The internal `RankedTemporalCandidate.candidate` field is also typed
as bare `object`, which erases the candidate's real type, so the function's
declared return type `list[CandidateT]` doesn't match what's actually
returned either (confirmed by mypy).

### Steps to reproduce

1. From the repo root, create a file `check_typecheck.py`:

   from dataclasses import dataclass
   from waggle.retrieval.rankers.temporal import rank_temporal

   @dataclass
   class NoTimestamp:
       text: str  # no `ts` field

   rank_temporal("query", [NoTimestamp("x")], direction="latest", top_k=1)

2. Run it directly:
   PYTHONPATH=src python3 check_typecheck.py

3. Separately, run mypy against the library file itself:
   PYTHONPATH=src python3 -m mypy src/waggle/retrieval/rankers/temporal.py \
       --ignore-missing-imports --python-version 3.12

### Expected behavior

- Step 2: calling rank_temporal with candidates that don't have `.ts` should
  either be rejected at type-check time, or fail with a clear, intentional
  validation error.
- Step 3: mypy should report zero errors for this file — the declared
  types should accurately describe what the function requires and returns.

### Actual behavior

- Step 2: the call type-checks fine under the current (unbound) TypeVar,
  then raises `AttributeError: 'NoTimestamp' object has no attribute 'ts'`
  at runtime with no earlier warning.
- Step 3: mypy reports:
  src/waggle/retrieval/rankers/temporal.py:67: error: "CandidateT" has no
    attribute "ts"  [attr-defined]
  src/waggle/retrieval/rankers/temporal.py:98: error: List comprehension
    has incompatible type List[object]; expected List[CandidateT]  [misc]

### Waggle version

0.1.25 (per pyproject.toml at time of report)

### Install method

None

### Environment

_No response_

### Operating System

Windows

_No response_

### Screenshot

N/A — this is a static type-checking finding (mypy), not a UI or CLI
runtime issue with visual output. Terminal output is pasted verbatim above
under "Actual behavior" and in Logs below in place of a screenshot.

### Logs

$ PYTHONPATH=src python3 -m mypy src/waggle/retrieval/rankers/temporal.py --ignore-missing-imports --python-version 3.12
src/waggle/retrieval/rankers/temporal.py:67: error: "CandidateT" has no attribute "ts"  [attr-defined]
src/waggle/retrieval/rankers/temporal.py:98: error: List comprehension has incompatible type List[object]; expected List[CandidateT]  [misc]
Found 2 errors in 1 file (checked 1 source file)

$ PYTHONPATH=src python3 check_typecheck.py
Traceback (most recent call last):
  File "check_typecheck.py", line 11, in <module>
    rank_temporal("query", [NoTimestamp("x")], direction="latest", top_k=1)
  File ".../src/waggle/retrieval/rankers/temporal.py", line 67, in <listcomp>
    timestamp=parse_timestamp(candidate.ts),
                               ^^^^^^^^^^^^
AttributeError: 'NoTimestamp' object has no attribute 'ts'

webgo-oss:
## Summary
Fixes #712 — `rank_temporal()` accepted candidates without a `.ts`
attribute due to an unbound generic type, causing a runtime
`AttributeError` instead of a static type error.

## Bug details
- **File:** `src/waggle/retrieval/rankers/temporal.py`
- **Function:** `rank_temporal()`

| # | Issue | Location |
|---|---|---|
| 1 | `CandidateT = TypeVar("CandidateT")` was unbound — nothing declared that candidates must have `.ts` | line 14 |
| 2 | `candidate.ts` accessed unconditionally with no static guarantee it exists → `AttributeError` risk at runtime for any caller passing the wrong shape | line 67 |
| 3 | `RankedTemporalCandidate.candidate` typed as bare `object`, erasing the candidate's real type | line 21 |
| 4 | Consequence of #3: the function's declared return type `list[CandidateT]` didn't actually match what was returned (mypy: `List comprehension has incompatible type List[object]; expected List[CandidateT]`) | line 98 |

## Fix
- Added a `HasTimestamp` Protocol requiring `ts: datetime | str`.
- Bound `CandidateT` to it: `TypeVar("CandidateT", bound=HasTimestamp)`.
- Made `RankedTemporalCandidate` generic (`Generic[CandidateT]`) with `candidate: CandidateT` instead of `object`.
- Updated the instantiation site to `RankedTemporalCandidate[CandidateT](...)`.

## Testing
- `python3 -m pytest tests/test_temporal.py -v` → all 4 existing tests pass unchanged (confirms no behavior change).
- Added a throwaway negative check: calling `rank_temporal` with a candidate missing `.ts` now fails at type-check time (`error: Value of type variable "CandidateT" of "rank_temporal" cannot be "NoTimestamp" [type-var]`), instead of only at runtime.
- `py_compile`, `pyflakes`, and `mypy --ignore-missing-imports` all pass clean on the changed file.

## Risk
None — purely a type-annotation change, no runtime logic touched. Existing callers with valid candidates are unaffected.

Closes #712

## Summary by CodeRabbit

* **Refactor**
  * Improved temporal ranking type safety and consistency.
  * Preserved candidate types through temporal ranking results.
  * Added timestamp requirements for supported candidates.

### DIFF
--- a/src/waggle/retrieval/rankers/temporal.py
+++ b/src/waggle/retrieval/rankers/temporal.py
@@ -1,24 +1,32 @@
 """Temporal ranking with topic gating for latest/oldest graph queries."""
 
 from __future__ import annotations
-
+# AFTER
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Literal, TypeVar
+from typing import Generic, Literal, Protocol, TypeVar
 
 from waggle.retrieval.scorers.topic_relevance import TopicScore, score_topic_relevance
 
 TOPIC_THRESHOLD = 0.35
 TemporalDirection = Literal["latest", "oldest"]
-CandidateT = TypeVar("CandidateT")
+
+
+class HasTimestamp(Protocol):
+    """Structural type for anything rank_temporal can sort by time."""
+
+    ts: datetime | str
+
+
+CandidateT = TypeVar("CandidateT", bound=HasTimestamp)
 
 
 @dataclass(frozen=True)
-class RankedTemporalCandidate:
+class RankedTemporalCandidate(Generic[CandidateT]):
     """A temporal candidate paired with its topic score."""
 
-    candidate: object
+    candidate: CandidateT
     topic_score: TopicScore
     timestamp: datetime
 
@@ -57,7 +65,7 @@ def rank_temporal(
         raise ValueError("direction must be 'latest' or 'oldest'.")
 
     ranked = [
-        RankedTemporalCandidate(
+        RankedTemporalCandidate[CandidateT](
             candidate=candidate,
             topic_score=score_topic_relevance(
                 query,


ard12 (changes requested):
Good catch on binding `CandidateT` to prevent the runtime `AttributeError`. Two small cleanups needed:

1. There's a leftover `# AFTER` comment on line 4 in `src/waggle/retrieval/rankers/temporal.py` that should be removed.
2. In `HasTimestamp`, `ts: datetime | str` is currently declared as a mutable attribute, which makes it invariant and can cause issues with frozen classes (like `Node.ts`). Defining it as a read-only property makes it covariant and plays much nicer with mypy:
```python
class HasTimestamp(Protocol):
    @property
    def ts(self) -> datetime | str: ...
```

Also, minor nit: the PR title mentions `rank_topics` instead of `rank_temporal`.

PR #714, opened 2026-09-01
perf(graph): batch single embed() calls during multi-node observation ingest (fixes #124)

### LINKED ISSUE #124: perf(graph): batch single embed() calls during multi-node observation ingest

turfin-logic:
Resolves #124 by batching embedding calls during multi-node ingestion and decomposition.

`add_node` accepts an optional precomputed embedding. Both batch paths validate the returned vector count and preserve the existing per-item fallback when batching fails or returns an invalid result.

The three trailing-whitespace lines that caused `Lint (ruff)` to fail with `W293` have been removed.

Validation:
- `ruff check src/ tests/`
- `ruff format --check src/ tests/`
- `python scripts/check_markdown_links.py`
- `python scripts/sync_version.py --check`

All listed checks pass locally.

## Summary by CodeRabbit

- **New Features**
  - Added topic result caching with automatic refresh when graph data changes, plus an option to force recomputation.
  - Expanded exports with filtering, date ranges, redaction, signing, dependency inclusion, and low-confidence edge controls.
  - Exported snapshots now include embeddings by default.

- **Improvements**
  - Improved conversation processing speed through batch embedding.
  - Preserved observed timestamps across repository and context-window operations.
  - Added automatic community refresh tracking after graph updates and retention cleanup.


turfin-logic:
The `ruff` failure is fixed: all three W293 trailing-whitespace violations were removed, and the full lint job commands pass locally. The refreshed CI and HOL workflow runs are currently waiting for maintainer approval.

ard12 (changes requested):
Thanks for working on this. Batching embeddings here is a solid win, but there are two parity gaps to address before we can merge:

1. In `src/waggle/neo4j_graph.py`, `decompose_and_store` still calls `add_node` in a loop with individual embedding calls. We should batch `atomic_items` there too so it matches SQLite's `decompose_and_store`.
2. Error/fallback handling: in `src/waggle/graph/transcript.py`, `observe_conversation` raises a `ValueError` when `len(_batch_embeddings) != len(_candidate_texts)`. In this PR, `neo4j_graph.py` silently falls back to `_batch_embeddings = None`. We should keep the behavior consistent across backends (either raise or fall back in both).

Once `decompose_and_store` in Neo4j is updated and the fallback handling aligns with the SQLite implementation, this should be good to go.

PR #715, opened 2026-09-02
test: cover deterministic sort order in merge_evidence_records

madhav1811:
## Summary
- Adds coverage for `merge_evidence_records` to lock in deterministic sort order on ties (observed_at/turn_index) and confirm Python's stable sort preserves input order for full ties, keeping exports byte-for-byte reproducible across runs.

## Test plan
- [x] `pytest tests/test_evidence.py`

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify deterministic ordering of evidence records when timestamps match.
  * Confirmed records are ordered by descending turn index, with fully tied records preserving their original order.

### DIFF
--- a/tests/test_evidence.py
+++ b/tests/test_evidence.py
@@ -117,6 +117,41 @@ def test_merge_handles_empty_inputs() -> None:
     assert merge_evidence_records([], []) == []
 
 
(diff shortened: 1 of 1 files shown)


ard12 (approved):
LGTM. Clean test coverage for the tiebreak behavior and verifies stable sort ordering for reproducible exports.

